PR #138, opened 2014-06-17
Better Date and Time Handlers

jahumes:
I recently came across a pull request (https://github.com/stympy/faker/pull/73) to the original faker that wasn't implemented, but that I thought would add benefit to ffaker. Essentially, this pull request will add two new modules to the API to make generating Dates and Times a little more structured than just the basic Ruby way of using random numbers.

For the most part, this is just an re-implementation of vitormil's code. If you have any suggestions or thoughts, please let me know. 

## Date
````ruby
# generates a Date object in the interval
Faker::Date.between("01/08/2012", "02/08/2012")
# random within 60 days past
Faker::Date.backward(60) 
# random within 30 days future
Faker::Date.forward(30)
````

## Time
````ruby
# generates a Time object in the interval
Faker::DateTime.between("01/08/2012", "02/08/2012")
# random within 60 days past
Faker::DateTime.backward(60) 
# random within 30 days future between the hours of 6:00 PM to 11:59 PM
Faker::DateTime.forward(30, :night)
````


rocknruby:
I close this PR for now, if this is still needed we can discuss.

PR #141, opened 2014-07-15
Fix travis build for 1.8.7

whatthewhat:
Another option is to remove 1.8.7 from `.travis.yml` :smile:


rocknruby:
@whatthewhat we dropped the support for 1.8.7 - https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/

PR #151, opened 2014-09-08
[feature] - Support currency code, name generate

MQuy:
I think we should add support for currency code and name to make our ffaker more flexible.
About data i copy from google https://developers.google.com/public-data/docs/canonical/currencies_csv
What's your opinion about that?


rocknruby:
@PapePathe where is the test for this feature? Did you review this? 

The method#rand is now deprecated we use Array#sample for now.

PapePathe:
ok will add tests to it!

PR #152, opened 2014-09-18
Add Cuban phone numbers

luilver:


lib/ffaker/phone_number_cu.rb
```diff
@@ -0,0 +1,133 @@
...
+    # Home or Work Operator prefixes
+    HomeWorkOperatorsPrefix = %w(021 022 023 024 031 032 033 041 042 043 045 046 047 048 07)
+
+    OperatorsPrefix = MobileOperatorsPrefix + HomeWorkOperatorsPrefix
+
+    # Return a prefix in MobileOperatorsPrefix
+    # @see Faker::PhoneNumberCU::MobileOperatorsPrefix
+    #
+    def mobile_phone_prefix
+      MobileOperatorsPrefix[rand(MobileOperatorsPrefix.size)]
```
rocknruby: #rand is deprecated

PapePathe: ok i will remove it!

PR #155, opened 2014-10-10
Stop support ruby1.8.7

onigra:
Ruby 1.8.7 is Dead (ﾉ-_-)ﾉ ~┻━┻

https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/


rocknruby:
+1 for dropping :+1:

onigra:
:smile:

PR #160, opened 2014-11-06
set up sport methods with data

BenMorganIO:
This adds sports with some data for ffaker. There's nearly 1000 sports. The positions are of hockey and football. The leagues are from hockey. Captains and coaches are from hockey. Locations lists the biggest arenas in the world.


rocknruby:
Hi @BenMorganIO , thank you for contribution. Could you please add tests for sport methods. 

:octocat:

itsdan:
Isn't it better to use FFaker::Name for captains and coaches?

PR #162, opened 2014-12-24
V1 canadian 

guinslym:
Adding +-180  canadian cities for ffaker::AddressCA. I did also add a spec def for the cities.


BenMorganIO:
:+1:

Issue #161, opened 2014-11-14
New maintainer needed


EmmanuelOga:
Hey guys,

I feel sorry for the state of this gem but I barely use it for anything anymore and it is hard for me to find time to maintain it. If you would like to step up and take the helm let me know.

Thanks!

P.S.: just to clarify the part I feel sorry about is the number of open issues and pull requests... the codebase itself is pretty clean and shouldn't be hard to mantain :smile:

EmmanuelOga:
@rocknruby jumped to help, thanks so much!

PapePathe:
+1 @rocknruby

PR #164, opened 2015-01-06
refactored the heart of "slug" to be about 40% faster

dannysperry:
FFaker::Internet.slug was using an unnecessary `.gsub` that I refactored out. 

```
                    user     system      total        real
with .gsub       0.000000   0.000000   0.000000 (  0.000011)
without .gsub    0.000000   0.000000   0.000000 (  0.000004)   
```

### DIFF
--- a/lib/ffaker/internet.rb
+++ b/lib/ffaker/internet.rb
@@ -71,7 +71,7 @@ module Faker
     def slug(words = nil, glue = nil)
       glue ||= SLUG_DELIMITERS.rand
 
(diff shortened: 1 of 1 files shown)


EmmanuelOga:
Wow this is what I call attention to detail :-) Anyway, thanks for the PR! I'm offering ppl that already contributed the chance to take care of the gem. Would you be interested? See #161 

Thanks!

rocknruby:
Good job @dannysperry :rocket:

PR #166, opened 2015-01-08
Improve KR

marocchino:
- Add JobKr, GenderKR, PhoneNumberKR
- Improve NameKr, Gender, GenderCN


lib/ffaker/name_kr.rb
```diff
@@ -6,7 +6,11 @@ module NameKR
     extend self
 
     def name
-      "#{first_name}, #{last_name}"
+      "#{last_name}#{first_name}"
```
rocknruby: Hi @marocchino, does it make sence that the name method will return last_name and firstname without a space?

marocchino: @rocknruby I think so. It is most common in Korea to write the name without a space.

rocknruby: If this is the common way, then it's okay!

lib/ffaker/gender_kr.rb
```diff
@@ -0,0 +1,17 @@
+# encoding: utf-8
+
+module Faker
+  module GenderKR
+    extend ModuleUtils
+    extend self
+
+    GENDERS = k %w(남 녀)
```
rocknruby: I'm wondering that this code is working and your test doesn't failed. What does the "k" mean? Maybe I missing some ruby knowledge. :octocat:

lib/ffaker/phone_number_kr.rb
```diff
@@ -0,0 +1,50 @@
...
+      Faker.numerify("#{HOME_PHONE_PREFIXES.rand} #### ####")
+    end
+
+    def mobile_phone_number
+      Faker.numerify("#{MOBILE_PHONE_PREFIXES.rand} #### ####")
+    end
+
+    def international_mobile_phone_number
+      number = mobile_phone_number
+      number[0] = "+82 "
```
rocknruby: Good moment to use your country_code method, right?

lib/ffaker/gender_kr.rb
EmmanuelOga: Ah, this is a class method that gets embedded in all modules:

https://github.com/EmmanuelOga/ffaker/blob/master/lib/ffaker/utils/module_utils.rb#L5

It "converts" the array to a frozen array, extended with a couple of useful methods like "random_pick"

https://github.com/EmmanuelOga/ffaker/blob/master/lib/ffaker/utils/array_utils.rb

The reason we freeze it is so you get an exception if you try to mutate the array (this could happen if one method in ffaker mistakenly returns the source array instead of picking a random element, for instance).

EmmanuelOga: Uh, now that I look at this, it seems like this code could use some updates...

random_pick may not be needed anymore (there's the Array#sample method now), and I'm pretty sure the shuffle is not correct/fair (it should be implemented with an algorithm like [Fisher-Yates](http://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle)

EmmanuelOga: Ha! there's also Array#shuffle for free :-) This definitely looks like we could get rid of ArrayUtils now :smile: :sparkles:

rocknruby: Ahh, I got it! :dancers:

rocknruby: Since we are not supporting 1.8.x, best-practice will be to use Array#sample instead of the ArrayUtils#rand method, right?

EmmanuelOga: Right. You could still leave the k method if you wanted, but it would need to be changed to something like:

```ruby
    def k(array)
      array = argument.is_a?(Array) ? argument : argument.to_a
      array.each { |e| e.freeze }
      array.freeze
      array
    end
```

... then you would need to search and replace random_pick to sample all around (not sure if the api is exactly the same). Seems like a hopefully easy but a big change!

marocchino: fix in 04bc4ca

lib/ffaker/phone_number_kr.rb
marocchino: fix in  04bc4ca

lib/ffaker/gender_kr.rb
marocchino: Replace all `Array#rand` to `Array#sample` in cfe39eb

lib/ffaker/internet.rb
```diff
@@ -33,8 +33,8 @@ def user_name(name = nil)
         when 0
           Name.first_name.gsub(/\W/, '').downcase
         when 1
-          parts = [ Name.first_name, Name.last_name ].each {|n| n.gsub!(/\W/, '') }
-          parts = parts.join ArrayUtils.rand(%w(. _))
+          parts = [ Name.first_name, Name.last_name ]
+          parts = parts.join(%w(. _).sample).gsub(/\W/, '')
```
rocknruby: This will not return the same result as the old code. 

```ruby
parts.join(%w(. _).sample).gsub(/\W/, '') # will remove also the "."
```

lib/ffaker/utils/array_utils.rb
```diff
@@ -11,10 +11,6 @@ def self.random_pick(array, n)
       indexes.map { |n| array[n].dup }
     end
 
-    def self.rand(array)
```
rocknruby: Lets keep the rand method until we release the next major version and lets append a deprecation warning to keep the community happy and informed. Dont know if someone uses this Utils#rand methods in the tests. What do you think @EmmanuelOga, @marocchino?

rocknruby:
Good job @marocchino and thanks for your collaboration. For the next big refactoring I would be even happier if you make smaller PRs or PRs where you just finish one task. It took some time to review this huge PR, anyway I had fun. :dancers:

marocchino:
You are right. I will make another PR today.

lib/ffaker/utils/array_utils.rb
marocchino: Fix in 06e0faa

lib/ffaker/internet.rb
marocchino: fix in 7b6e179

lib/ffaker/phone_number_kr.rb
```diff
@@ -0,0 +1,54 @@
...
+      when 0 then international_mobile_phone_number
+      when 1 then international_home_work_phone_number
+      end
+    end
+
+    def country_code
+      "+82"
+    end
+
+    HOME_PHONE_PREFIXES = %w(02 031 032 033 041 042 043 044 049 051 052 053 054 055 061 062 063 064)
```
rocknruby: I guess this is my last comment :+1: , I'm following the ruby style guide from @bbatsov - https://github.com/bbatsov/ruby-style-guide and think it useful to keep the constants before defining methods, as you can see here: https://github.com/bbatsov/ruby-style-guide#classes--modules

marocchino: Totally right, I fix some other style issue in  5882f6c.

rocknruby:
Hi @marocchino, could you rebase and squash commits to keep a intelligible master branch. Afterwards I'll merge and release a new version :+1:

marocchino:
I squash it. Thank you. :+1:

PR #167, opened 2015-01-09
Update travis.yml

teeparham:
* cache and sudo settings will speed up travis builds
* Specify 2.1 and 2.2 to test the latest patch versions of 2.1 and 2.2,
which is sufficient.


rocknruby:
that makes sense :+1:

PR #188, opened 2015-04-15
adding sports to faker modules

fbenevides:


test/test_sports.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+
+require 'helper'
+
+class TestMovie < Test::Unit::TestCase
```
rocknruby: can you please rename your class :lollipop:

fbenevides: done.

PR #195, opened 2015-05-04
Ensure that PhoneNumber does not throw invalid US area codes

rstacruz:
Related: #181 

This fixes issues with `FFaker::PhoneNumber.phone_number` generating numbers with area codes that begin with 0. According to [this list](https://en.wikipedia.org/wiki/List_of_North_American_Numbering_Plan_area_codes) (wikipedia.org), area codes beginning with 0 and 1 are reserved.


marocchino:
:+1:

rstacruz:
self-merging, hope no one minds.

PR #201, opened 2015-05-12
Improve some tests

marocchino:


test/test_sports.rb
```diff
@@ -3,7 +3,11 @@
 require 'helper'
 
 class TestSports < Test::Unit::TestCase
+  def setup
+    @tester = FFaker::Sport
+  end
+
   def test_name
-    assert_match /[ a-z]+/, FFaker::Sport.name
```
marocchino: This test fails when name is a "BMX".

PR #203, opened 2015-05-12
Deprecate random_pick

marocchino:
Because [`Array#sample`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-sample) can take argument.

### DIFF
--- a/lib/ffaker/vehicle.rb
+++ b/lib/ffaker/vehicle.rb
@@ -26,7 +26,7 @@ module FFaker
 
     def manufacturer_color(n=2)
       # Take two prefixes because it's more fun than one
-      [COLOR_PREFIXES.random_pick(n), self.base_color].flatten.join(' ')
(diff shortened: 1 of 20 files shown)


rstacruz:
:+1: on this from me, always found that pretty strange

lib/ffaker/vehicle.rb
```diff
@@ -26,7 +26,7 @@ def make
 
     def manufacturer_color(n=2)
       # Take two prefixes because it's more fun than one
-      [COLOR_PREFIXES.random_pick(n), self.base_color].flatten.join(' ')
+      [COLOR_PREFIXES.sample(n), self.base_color].flatten.join(' ')
```
rstacruz: probably better as `(COLOR_PREFIXES.sample(n) + [self.base_color]).join(' ')`, but for another time

marocchino:
@rstacruz 's review is applied.
I removed all ArrayUtils related code in libs. :)
Hopefully, We can remove ArrayUtils next update.

PR #205, opened 2015-05-15
Remove deprecation messages

itsdan:
UK & US address were deprecated ~30 versions ago.


rstacruz:
This file was automatically generated using `rake reference`... might want to look at that script instead. Thanks for the PR, but running the rake task again (which will probably happen when new methods get implemented) will just override these again.

itsdan:
Hello, @rstacruz! What about removing UK & US methods from those modules? They were extracted a long time ago (4 years).

itsdan:
ping @rstacruz

rstacruz:
that idea gets a :+1: for me but I'd rather see what the feedback of other maintainers are on this as well :)

Issue #190, opened 2015-05-01
Get rid of Rake task :gemspec


rstacruz:
https://github.com/EmmanuelOga/ffaker/commit/c84202974c9e1c9f9235780bcf3c01cb2648b9c1

There's no need to dynamically generate gemspec files.

``` rb
Gem::Specification.new do |s|
  s.name              = 'ffaker'
  s.version           = '2.0.0'
  s.rubyforge_project = 'ffaker'
  s.required_ruby_version = '>= 1.9'

  s.license = 'MIT'

  s.summary     = "Faster Faker, generates dummy data."
  s.description = "Faster Faker, generates dummy data."

  s.authors  = ["Emmanuel Oga"]
  s.email    = 'EmmanuelOga@gmail.com'
  s.homepage = 'http://github.com/emmanueloga/ffaker'

  s.require_paths = %w[lib]

  s.rdoc_options = ["--charset=UTF-8"]
  s.extra_rdoc_files = %w[README.md LICENSE Changelog.md]

  s.files = `git ls-files`.strip.split("\n") # <---
  s.test_files = s.files.select { |path| path =~ /^test\/test_.*\.rb/ }

  s.add_development_dependency 'rake', '~> 10.1.1'
  s.add_development_dependency 'test-unit'
end
```

PR #206, opened 2015-05-15
Migrate from Test::Unit to Minitest

itsdan:
Use Ruby's standard test library.


marocchino:
I am OK with minutest but,
- You should change gemspec file also.
- And, It's failing on travis.

itsdan:
Hello, @marocchino!
Thanks for the reply! I've updated .gemspec file. It's now passing on Travis (except for 1.9.3). And maybe it's time to drop 1.9.3 support: if new languages (Address, Name) are added to the gem, files will have to contain encoding comment at the top.

rstacruz:
why is @marocchino not a collaborator yet? :dancer:

rstacruz:
also :+1: for getting rid of that long list of files in the gemspec. #190 would need to update the `rake :build` task too, i think

PR #214, opened 2015-06-26
chinese phone support

borisding1994:


rocknruby:
@poyueh I'm missing specs, I will be much happier, if you could add some :+1:

PR #216, opened 2015-07-12
Improve NameJA

suiyujin:
##### Changes
* FFaker::NameJA.name generate name in the order of last_name, first_name.
* Add test_first_name, test_last_name and test_last_first.
* test_last_first and test_name examine the order of name.

###### FFaker::NameJA.name generate name in the order of last_name, first_name.

Japanese name is not written in order of first_name, last_name like Korean.
I refer to this commit (https://github.com/ffaker/ffaker/commit/335691c16210904951129e66142725abf3c514f4#diff-9144e4d488890fffa372d862fc110957). 
This commit defined last_first as an alias of name.
But I think name should be an alias of last_first, because it is natural.

###### test_last_first and test_name examine the order of name.
https://github.com/suiyujin/ffaker/blob/improve-japanese-name/test%2Ftest_name_ja.rb#L6-16
This section replaces all data to small data, to avoid slowing down.

### DIFF
--- a/lib/ffaker/name_ja.rb
+++ b/lib/ffaker/name_ja.rb
@@ -5,10 +5,6 @@ module FFaker
     extend ModuleUtils
     extend self
 
-    def name
-      "#{first_name}#{last_name}"
-    end
-
     def first_name
       FIRST_NAMES.sample
     end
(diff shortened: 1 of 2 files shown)


rstacruz:
I'm not Japanese but I trust you know what you're doing. Thanks! ;)

rocknruby:
:+1:

PR #218, opened 2015-07-27
Adds basic FFaker::Address.time_zone support

dugjason:


test/test_address.rb
```diff
@@ -83,4 +83,8 @@ def test_country_code_of_particular_country
     assert_match 'UA', FFaker::Address::country_code('Ukraine')
     assert_match /[A-Z]{2}/, FFaker::Address::country_code('Foo')
   end
+
+  def test_time_zone
+    assert_match /[ a-z]+/, FFaker::Address::time_zone
```
marocchino: This test is pointless.
How about `assert_include FFaker::Address::TIME_ZONE, FFaker::Address. time_zone` or `/\A[a-z_\/]\z/i`?

dugjason:
Thanks for the review @marocchino. I've added both the ```assert_include FFaker::Address::TIME_ZONE``` comparison, and a more sensible regex.

marocchino:
@dugjason Thanks, but we don't need both because It's too verbose. >_< Could you choose one of them please?

dugjason:
Sure thing. I just included both examples for completeness, but I see it's one of the only tests with more than one test case.

marocchino:
:+1:

PR #219, opened 2015-08-02
Vin generator result now does not include I, O or Q letters

fabrouy:
As per: https://en.wikipedia.org/wiki/Vehicle_identification_number

*One consistent element of the VIS is the 10th digit, which is required worldwide to encode the model year of the vehicle. Besides the three letters that are not allowed in the VIN itself (I, O and Q), the letters U and Z and the digit 0 are not used for the model year code. Note that the year code is the model year for the vehicle.*

There's more work to do to ensure that generated VIN codes are valid and not only "format valid".

### DIFF
--- a/test/test_vehicle.rb
+++ b/test/test_vehicle.rb
@@ -32,7 +32,7 @@ class TestVehicle < Test::Unit::TestCase
   end
 
   def test_vin
-    assert_match /\A[A-Z0-9]{17}\z/, FFaker::Vehicle.vin
+    assert_match /\A[A-Z0-9&&[^IOQ]]{17}\z/, FFaker::Vehicle.vin
   end
 
   def test_drivetrain
--- a/lib/ffaker/vehicle.rb
+++ b/lib/ffaker/vehicle.rb
@@ -11,6 +11,7 @@ module FFaker
       beautiful bright calm dangerous dark dull fast magnetic magnificent majestic melodic metallic
       mundane mute mysterious new pleasant pretty resonant royal slate soft tranquil vibrant weak
     ))
+    VIN_LETTERS = (LETTERS - %w(i o q)).map { |letter| letter.upcase }
 
     def base_color
       FFaker::Color.name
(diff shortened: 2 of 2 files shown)


test/test_vehicle.rb
```diff
@@ -32,7 +32,7 @@ def test_trim
   end
 
   def test_vin
-    assert_match /\A[A-Z0-9]{17}\z/, FFaker::Vehicle.vin
+    assert_match /\A[A-Z0-9&&[^IOQ]]{17}\z/, FFaker::Vehicle.vin
```
marocchino: This test seems not to working.
how about this?
```ruby
/\A1\d/[A-HJ-NPR-Z]{3}\d{5}[A-HJ-NPR-Z]\d{6}\z/
```

fabrouy: Did you run it many times? It fails sometimes with:

`for i in {1..100}; do bundle exec ruby -Itest test/test_vehicle.rb; done`

marocchino: It was not happen to me. -3-
Can I see your fail message?

fabrouy: This is what I get running the test I wrote without my code update.

ffaker

marocchino:
but, It's your regexp not mine.

test/test_vehicle.rb
marocchino: but, It's your regexp not mine.

fabrouy: I'm not following.
You told me that my test was not working and you submitted another regex, so I'm showing you that with my regex the original code (before my patch) fails. 

What is the reason you say my test is not ok?

marocchino: that because, your regexp works like this. :neutral_face: 

[screen shot 2015-08-14 at 8 53 37 am](http://regexper.com/#%2F%5CA%5BA-Z0-9%26%26%5B%5EIOQ%5D%5D%7B17%7D%5Cz%2F)

rocknruby:
I'll close this PR because of unstable tests. If you fix them @fabrouy and I guess @marocchino would help you dont hesitate to open another PR.

PR #221, opened 2015-08-06
Exclude invalid area codes

kevintraver:


rstacruz:
Thanks fro the contribution, but I'm not sure it's a good idea to implement it this way... you will be doing `Range#to_a` of a pretty big range every time you call the function.

kevintraver:
What about just using the original range, but then trying again it is invalid?

kevintraver:
@rstacruz let me know your thoughts

marocchino:
@kevintraver Thanks! It needs to clean up little bit but, good enough to merge :3

lib/ffaker/phone_number.rb
```diff
@@ -30,7 +30,12 @@ def area_code
       # The North American Numbering Plan (NANP) does not permit the digits 0
       # and 1 as the leading digit of an area code.
       # https://en.wikipedia.org/wiki/List_of_North_American_Numbering_Plan_area_codes
-      rand(201..999)
+      rand_area_code = rand(201..999)
+      invalid_area_codes = [211,311,411,511,611,711,811,911]
+      while(invalid_area_codes.include?(rand_area_code))
+        rand_area_code = rand(201..999)
+      end
```
rstacruz: this works too:

```rb
do
  rand_area_code = ...
while invalid_area_codes.include?(rand_area_code)
```

marocchino: eda18d1 :3

PR #222, opened 2015-08-08
Add new country - Ukraine, cover it with tests, fix some bugs

kovpack:
Changes:
 - added support **of new country - Ukraine**;
 - all methods covered with tests;
 - correction of small mistakes.

@rstacruz , @marocchino , I've got a few notes:

1. `numerify` method has one bug and can not be applied to all cases when there is necessity to numerify some mask: in many cases it generates numbers like `098` that can not be applied (at least in Ukraine) to apartment numbers, street numbers etc. I added private method to support Ukrainian properly and did not touch original method (touching it requires a lot of changes).

2. `const_missing` is buggy and not thread-safe. When I was checking data with regexp in tests I found out, that in approximately 50% of all cases it assigned English data instead of Ukrainian. So 100% of tests that look like `assert FFaker::AddressSN::REGIONS.include?(FFaker::AddressSN.region)` or `assert_include @tester::FIRST_NAMES_FEMALE, (@tester.first_name_female)`, actually, do not test anything and simply are useless. Someone has to implement thread-safe `const_missing` method or just move data files that has more than 50-100 elements into `.rb` files, convert them to arrays and assign directly to the constants. It's sad that current implementation does not work properly, cause the idea to load constants dynamically and put data to separate files - is really nice.


kovpack:
@rstacruz , @marocchino it would be really grateful if someone merges this to include in the nearest release. I'd like to use this in my current project :)

lib/ffaker/address_ua.rb
```diff
@@ -0,0 +1,61 @@
...
...шаллові Острови", "Мексика", "Мозамбік", "Молдова", "Монако", "Монголія", "Намібія", "Науру", "Непал", "Нігер", "Нігерія", "Нідерланди", "Нікарагуа", "Німеччина", "Нова Зеландія", "Норвегія", "Об’єднані Арабські Емірати", "Оман", "Пакистан", "Палау", "Панама", "Папуа-Нова Гвінея", "Парагвай", "Перу", "Південна Корея", "Південний Судан", "Південно-Африканська Республіка", "Північна Корея", "Польща", "Португалія", "Російська Федерація", "Руанда", "Румунія", "Сальвадор", "Самоа", "Сан-Марино", "Сан-Томе і Принсіпі", "Саудівська Аравія", "Свазіленд", "Сейшельські Острови", "Сенеґал", "Сент-Вінсент і Гренадини", "Сент-Кітс і Невіс", "Сент-Люсія", "Сербія", "Сирія", "Сінгапур", "Словаччина", "Словенія", "Соломонові Острови", "Сомалі", "Судан", "Суринам", "Східний Тимор", "США", "Сьєрра-Леоне", "Таджикистан", "Таїланд", "Танзанія", "Того", "Тонга", "Тринідад і Тобаго", "Тувалу", "Туніс", "Туреччина", "Туркменістан", "Уганда", "Угорщина", "Узбекистан", "Україна", "Уругвай", "Федеративні Штати Мікронезії", "Фіджі", "Філіппіни", "Фінляндія", "Франція", "Хорватія", "Центральноафриканська Республіка", "Чад", "Чехія", "Чилі", "Чорногорія", "Швейцарія", "Швеція", "Шрі-Ланка", "Ямайка", "Японія"]
```
marocchino: Is there any reason to not extract this?
I think this should be data.

kovpack: Yeap, there is a reason - point 2) above in the description of pull request.

kovpack: I'll move it to proper place when `const_missing` method will be fixed. But with current implementation tests simply do not pass or pass randomly.

kovpack: @marocchino , here is what I mean telling that in case of moving big chunks of data into `data` folder constants are simply assigned in a wrong way: it checked Ukrainian regular expression against English data. So most tests that check something like `assert_include @tester::COUNTRY, @tester.country` in 50% of cases check that some value from `COUNTRY` is equal to the result of `country` method, however, when this test applies, e.g. to Ukrainian, Russian or any other language different from English, it can actually check English names and not the needed ones :) I suppose, bugginess percent depends on the amount of cores and speed of the CPU, but on my MBP Retina 15 tests just fail randomly. That is why to skip using `const_missing` method I had to move countries to the file as an array.
```
...
Loaded suite /Users/kovpack/.rvm/gems/ruby-2.2.0/gems/rake-10.1.1/lib/rake/rake_test_loader
started
...................................................................................F
======================================================================================================================================================================================================
Failure:
  </[а-яА-ЯіїєґІЇЄҐ’\-\s]/> expected to be =~
  <"Greece">.
test_country(TestAddressUA)
/Users/kovpack/ruby_apps/ffaker/test/test_address_ua.rb:16:in `test_country'
     13:   end
     14:
     15:   def test_country
  => 16:     assert_match /[а-яА-ЯіїєґІЇЄҐ’\-\s]/, @tester.country
     17:   end
     18:
     19:   def test_province
```

marocchino: how about remove `include FFaker::Address`?

kovpack: yeap, that should help. i just did not want to copy-paste code like `time_zone` etc, however, copy-paste in this case will be less evil :) i'll try and will push changes in an hour or two

test/test_address_ua.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_country
+    assert_match /[а-яА-ЯіїєґІЇЄҐ’\-\s]/, @tester.country
+  end
+
+  def test_province
+    assert_match /[а-яА-ЯіїєґІЇЄҐ’\-\s]/, @tester.province
+  end
+
+  def test_zip_code
+    assert_match /^\d{5}$/, @tester.zip_code
```
marocchino: I thinks we need to test whole string with `\A` and `\z`
```ruby
"111
a".match(/^\d{3}$/)
```

kovpack: you are right, i've already noticed this and in the process of fixing it :)

kovpack:
@marocchino , I've refactored everything I wanted: fixed regular expressions for tests, moved data to data files etc.

By the way, I've also fixed regular expression for NameBR (it did not correspond its data and tests failed in some cases).

And I've also found bugs in AddressRU tests - ALL of them are useless and buggy :) But correcting that much is out of scope of this pull request, I suppose. There is the same problem as with Ukrainian. And the author of tests added `/\A[a-zа-я\-\s]+\z/i` as wrong workaround (so this expression works, because he added checking of English characters, that can not be present in Russian words), but when constant is assigned in some cases properly, tests fail, cause there is one more character, that is specific to Russian - `ё` (it should be added to regexp separately and `a-z` should be removed). Tests should be run multiple times to reproduce this. Failure looks like this:

```
  </\A[a-zа-я\-\s]+\z/i> expected to be =~
  <"Орёл">.
test_city(TestAddressRU)
/Users/kovpack/ruby_apps/ffaker/test/test_address_ru.rb:12:in `test_city'
      9:   end
     10:
     11:   def test_city
  => 12:     assert_match /\A[a-zа-я\-\s]+\z/i, @tester.city
     13:   end
     14:
     15:   def test_province
```

But in general, everything that related Ukrainian - is fixed.

lib/ffaker/data/address_ru/province
```diff
@@ -31,7 +31,6 @@
 Костромская область
 Краснодарский край
 Красноярский край
-Крым
```
ovcharik: Political github, lol.

kovpack: @ovcharik , have you ever heard of international and Ukrainian laws? If you did not - that is your problem, I suppose, and of the guy who added this information there. Just for your information - https://en.wikipedia.org/wiki/United_Nations_General_Assembly_Resolution_68/262
If some guy at night penetrates into neigbour's house, kills the owner and his wife with children, puts the house on fire, and steals neigbour's money and car from garage - does not make that car his own property. He'll drive it some time, of course, but soon he'll be caught and punished. Everything in accordance with the laws. Hope, my analogy is quite clear and descriptive and not "political" at all - I just fixed bug (hope, that was bug introduced by some child who watched too much of Russian TV and did not go to school). And, by the way, if you are interested to what country belongs, let us say, Crimean Sevastopol, I can give contacts of my colleague, who is from Sevastopol (works with me in pair) - he'll explain you everything in details, but not here.

kovpack:
@marocchino , I've also added Ukrainian lorem - very important thing. Functions pretty good :)

marocchino:
:+1:

admsev:
:+1:

PR #231, opened 2015-09-15
Stricter regexps

marocchino:


test/test_venue.rb
```diff
@@ -4,6 +4,6 @@
 
 class TestVenue < Test::Unit::TestCase
   def test_name
-    assert_match /1\+|[ a-z]+/i, FFaker::Venue.name
```
marocchino: ref : https://github.com/ffaker/ffaker/blob/master/lib/ffaker/data/venue/venue_list
there are no number in data.

PR #236, opened 2015-09-22
Tests for DA

marocchino:


lib/ffaker/name_da.rb
```diff
@@ -11,49 +11,29 @@ module NameDA
...
-      when 1..2 then "#{fname} #{last_name} #{last_name}"
-      else           "#{fname} #{last_name}"
-      end
+      name(:female)
     end
 
-    def first_name gender = :any
+    def first_name(gender = :any)
       case gender
-      when :any then (rand(2) == 0) ? name(:male) : name(:female)
```
marocchino: It was a full name. so, `any_name`, `first_name(:any)` and `name(:any)` could return a wrong string.

rocknruby:
good fix and it looks much more readable :+1:

PR #237, opened 2015-10-13
Indian addresses

ramkumar-kr:
This commits generates data for
- Indian states
- Union territories
- Pincodes
- Time zone

The FFaker::AddressIN class contains the methods to generate Indian
Addresses


lib/ffaker/address_in.rb
```diff
@@ -0,0 +1,56 @@
...
+    def state_and_union_territory_abbr
+      (STATE_ABBR + UNION_TERRITORY_ABBR).sample
+    end
+
+    def country
+      FFaker::Address.country("IN")
+    end
+
+    def country_code
+      FFaker::Address.country_code("India")
```
gustavor-souza: I don't think it is a country code, looks like you change things.

test/test_address_in.rb
```diff
@@ -0,0 +1,50 @@
+# encoding: utf-8
+
+require 'helper'
+
+class TestAddressININ < Test::Unit::TestCase
+  def test_in_state
+    assert_match /[ a-z]/, FFaker::AddressIN.state
```
gustavor-souza: /\A[A-Z][ A-Za-z]*\z/

PR #238, opened 2015-10-15
fix slug with params

marshallds:
Fixes a bug in Internet::slug where providing input words would not result in a valid slug.

Also adds a test for a specifying a glue symbol which was working but not tested.


marocchino:
Thanks!

PR #241, opened 2015-11-02
Use HTTPS endpoint of robohash.org

wouterw:
HTTP endpoint of Robohash was redirecting to HTTPS.
`open-uri` doesn't automatically follow redirects:

```
irb(main):001:0> require "open-uri"
=> true
irb(main):002:0> open("http://robohash.org/test")
RuntimeError: redirection forbidden: http://robohash.org/test -> https://robohash.org/test
irb(main):003:0> open("https://robohash.org/test")
^[[D=> #<Tempfile:/var/folders/bd/09v89b6j65g0k05wf0rsct5m0000gn/T/open-uri20151102-73867-1j19io5>
```

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -480,7 +480,7 @@
 
 | Method | Example |
 | ------ | ------- |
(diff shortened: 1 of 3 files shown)


rstacruz:
thanks!

PR #244, opened 2015-11-12
Refactor ul_short and ul_long methods to increase CC score

maxehmookau:
I was curious to see if I could improve the code climate score of the repo slightly by refactoring. It was mainly a personal exercise but thought I'd offer it back if you're interested.

I'm not entirely convinced what it adds in DRYness makes up for a slight drop in readability?
Either way, thanks for the gem, I've found it very helpful.

### DIFF
--- a/lib/ffaker/html_ipsum.rb
+++ b/lib/ffaker/html_ipsum.rb
@@ -31,35 +31,19 @@ module FFaker
     end
 
     def ul_short(items = 3)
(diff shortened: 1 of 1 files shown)


rocknruby:
I took your advice and motivation to generate a private function called `#content_tag_for` the naming based on rails helper method. You see it in PR #275. Thanks.

PR #246, opened 2015-11-23
[feature] add AddressGR

falegk:
Add Greek addresses

* Cities
* Regions
* Street names, numbers and street address
* zip code


lib/ffaker/address_gr.rb
```diff
@@ -0,0 +1,40 @@
...
+    def city
+      CITY.sample
+    end
+
+    def street_name
+      STREET.sample
+    end
+
+    def street_nbr
+      FFaker.numerify STREET_NUMBER.sample
```
marocchino: We don't need `STREET_NUMBER.sample` here.
```ruby
FFaker.numerify "# ## ###"
```
will be fine.

test/test_address_gr.rb
```diff
@@ -0,0 +1,37 @@
+# encoding: utf-8
+
+require 'helper'
+
+class TestAddressGR < Test::Unit::TestCase
+
+  def setup
+    @addressGr = FFaker::AddressGR
```
marocchino: agree with that. 
It is not common in ruby.

lib/ffaker/address_gr.rb
falegk: This returns a number that looks like: "3 45 678"
But we want to generate a number up to three digits. ( 5 or 45 or 345 )
I think that it's correct :smile:

marocchino: Oh. You are right! >_<

marocchino:
merged with 5667f13 ff9ad27 .

Thanks!

PR #252, opened 2016-01-31
Airlines names and flight numbers generator

jakubm:
New Airline generator with the following functionality
- Airline names
- Flight numbers (number portion between 1 and 3999

Both datasets taken from Prague Airport traffic data for the last 3 years.

Also fix for one test (test address in)


marocchino:
Thanks!

PR #255, opened 2016-02-19
US phone exchange codes follow area code rules

mgomes:
US exchange codes have to follow the same rules as area codes: https://en.wikipedia.org/wiki/North_American_Numbering_Plan#Numbering_system. This PR makes that change.

I utilized the existing `area_code` generator, but let me know if you want me re-factor further. I also have a regex that has all of this phone logic built in that I can add to the specs. Let me know.


marocchino:
Thanks!

PR #258, opened 2016-02-29
[feature] - Support phone code in phone number module

MQuy:
I have a problem when creating a phone code for user model. After trying and looking at ffaker code, unfortunately, ffaker doesn't support phone code now. 
I think we should add this feature because it make user easier to generate phone code correctly.

Here is my suggest about it. What's your opinion?


marocchino:
Thanks!

Issue #256, opened 2016-02-23
Date Between


chrishough:
Have we considering adding the between for dates like Faker?
```

# Random date between dates
Faker::Date.between(2.days.ago, Date.today) #=> "Wed, 24 Sep 2014"

# Randome date between dates except for certain date
Faker::Date.between_except(1.year.ago, 1.year.from_now, Date.today) #=> "Wed, 24 Sep 2014"

# Random date in the future (up to maximum of N days)
Faker::Date.forward(23) # => "Fri, 03 Oct 2014"

# Random date in the past (up to maximum of N days)
Faker::Date.backward(14) #=> "Fri, 19 Sep 2014"
```

PapePathe:
+1 for this.

brnrdog:
I opened a PR to cover this feature (#263), I'm just waiting for a response.

rocknruby:
@chrishough #between is now implemented and just merged into the master with #263 - I close this issue right now.

PR #263, opened 2016-04-08
Add between method to FFaker::Time

brnrdog:
Add `between` method to `FFaker::Time` to get a random time within a time range.

It's an interesting feature and it was asked in #256


IgorMarques:
I believe you should add a entry here: https://github.com/ffaker/ffaker/blob/master/REFERENCE.md

rocknruby:
I like it. Thanks.

jeremymoritz:
This does not appear to be working correctly as documented in Rails 5.  

`
FFaker::Time.between(2.days.ago, Date.today)

NoMethodError: undefined method `to_f' for Fri, 06 Jul 2018:Date
Did you mean?  to_s
from ~/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/ffaker-2.8.1/lib/ffaker/time.rb:44:in 'between'`

PR #269, opened 2016-04-14
Prevent time generator to generate time with minutes equals to 60

IgorMarques:
This PR fix in an issue that could happen with ```FFaker::Time``` generator.

Sometimes this line:

```ruby
minutes = params[:minutes] || (rand * 60).ceil
```

Would allow the generation of minutes equals to 60 and that would end up throwing an ```ArgumentError: argument out of range```when generating dates at this line here:

```ruby
series = [date = ::Time.local(year, month, day, hours, minutes)]
```

## PS: issue with the failing build is addressed on this other PR: #266 ##

### DIFF
--- a/lib/ffaker/time.rb
+++ b/lib/ffaker/time.rb
@@ -31,7 +31,7 @@ module FFaker
 
     def datetime(params = {})
       hours = params[:hours] || (rand * 12).ceil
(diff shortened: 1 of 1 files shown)


rocknruby:
good fix 🚀

PR #278, opened 2016-06-04
FFaker Youtube

andyklimczak:
Add FFaker methods to create some youtube urls.

The 50 youtube video ids are the top 50 most viewed videos (as of today). I did this to remove possible bias, and these videos have a lower possibility of being removed.

My use case: In my project, I have image posts where people upload images, and video posts with links to youtube urls. 
So instead of having all of the seeded video posts be hard coded Taylor Swift Blank Space, I wanted to contribute to FFaker to help me out and provide some more interesting data.


rocknruby:
🚀

alepore:
I was believing that those were random generated, 11-chars ids... 
With just 50 ids it's super easy to get the same id, not very useful when using factories.

andyklimczak:
Randomly generating 11char ids would not be guaranteed to be valid youtube ids

Using actual video ids (from the top 50 videos) ensured that they were real videos, and have a low likelihood of being removed from youtube (and generally SFW). But yeah, I totally agree that only having 50 choices increases chance for duplicates. But that is the trade off if you want actual video ids, not just ids that look like youtube video ids.

Perhaps there is a chance for PR of top 100/500/1000 video ids?

PR #284, opened 2016-07-16
Swap test_address_ch_de.rb and test_address_ch_fr.rb contents to correctly match filename.

xunker:
While working on something else today, I found that `test_address_ch_de.rb` is testing `AddressCHFR` and `test_address_ch_fr.rb` is testing `AddressCHDE`. This pull swaps their contents.


marocchino:
👍 Thanks!

PR #288, opened 2016-07-29
Add Book module to FFaker

max-si-m:
Includes title, author, isbn, cover, genre, and description for books. It may be helpful for someone like me :)


marocchino:
Thanks!

PR #289, opened 2016-08-05
Add html_safe_name and html_safe_last_name

dimroc:
Guarantees returned names don’t have an apostrophe, which often breaks controller and view specs because they get escaped to &#39;. The return on test coverage for apostrophed names in this scenario is negligible.


marocchino:
LGTM.

PR #291, opened 2016-08-22
Generate file list automatically

marocchino:
- Remove tests
- will fix #280

```bash
$ gem build ffaker.gemspec
WARNING:  description and summary are identical
WARNING:  pessimistic dependency on rake (~> 10.1.1, development) may be overly strict
  if rake is semantically versioned, use:
    add_development_dependency 'rake', '~> 10.1', '>= 10.1.1'
WARNING:  open-ended dependency on test-unit (>= 0, development) is not recommended
  if test-unit is semantically versioned, use:
    add_development_dependency 'test-unit', '~> 0'
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: ffaker
  Version: 2.2.0
  File: ffaker-2.2.0.gem
$ gem install ffaker-2.2.0.gem
Successfully installed ffaker-2.2.0
1 gem installed
$ irb
[1] pry(main)> require 'ffaker'
=> true
[2] pry(main)> FFaker::IdentificationBR.cpf
=> "50027861388"
```

PR #293, opened 2016-09-10
Add more education degrees

tejasbubane:
Move `degree_prefix` and `degree_short_prefix` to data directory


marocchino:
Thanks!

tejasbubane:
@marocchino I just saw https://github.com/ffaker/ffaker/blob/master/ffaker.gemspec and seems like I need to add these new files to the list so that they are available in the gem bundle. Any reason why we are not using something like this?

```ruby
s.files         = Dir['lib/**/*' ........] 
```

marocchino:
That fixed in #291 but, Need some confirm.

PR #294, opened 2016-09-26
Update README.md

zedtux:
As on rubygems.org the Ruby version is not correctly set and allow a Ruby 1.8.7 developer to install an incompatible version, the README.md states now what version of the gem needs to be installed in order to work properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,16 @@ and then
 
     require 'ffaker'
 
+## Ruby  1.8.7
+
+FFaker 1.31.0 is the last version that works with legacy Ruby versions. This version is no longer supported; recent releases will only target recent Ruby versions (See [our Travis file](https://github.com/ffaker/ffaker/blob/master/.travis.yml) for the supported Ruby versions).
+
+```
+gem 'ffaker', '1.31.0'
+```
+
+Finally, you'll have to use the Faker class (v1) instead of FFaker (v2 and later).
+
 ## Why ffaker?
 
 ffaker is a fork of faker, and was initially written in an effort to speed up


README.md
```diff
@@ -28,6 +28,16 @@ and then
...
 
+## Ruby  1.8.7
+
+In the case you're still using Ruby 1.8.7, you cannot use the last version of ffaker:
+
+```
+gem 'ffake', '1.31.0'
+```
+
+Fianlly you will have to use the `Faker` class in order to generate data (instead of `FFaker` since the version 2).
```
tejasbubane: *`Finally`

README.md
```diff
@@ -28,6 +28,16 @@ and then
 
     require 'ffaker'
 
+## Ruby  1.8.7
+
+In the case you're still using Ruby 1.8.7, you cannot use the last version of ffaker:
+
+```
+gem 'ffake', '1.31.0'
```
tejasbubane: *`ffaker`

tejasbubane:
@zedtux Please add [`[ci skip]`](https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build) to the commit message for README-only changes so that the `CI` does not run unnecessarily.

zedtux:
Thank you @tejasbubane for the review and the comment.

marocchino:
I am not sure on this. 🤕 
The ruby 1.8.7 is retired and we drop it because we do not want our user use it.
If we add this phrases we should mention recommendation of ruby version up.

@ffaker/team can we discuss about it?

ref
- https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/
- #155

zedtux:
> ... we do not want our user use it.

Well why would you care what "your users" uses? It's not like you have troubles with Ruby 1.8.7, it's only about the dependencies.
As you said, just mention that you don't support it and this is just as information and you're done.

tejasbubane:
> just mention that you don't support it and this is just as information and you're done

I second this.

marocchino:
Well, I should apology for misunderstandings on this. But It is not that simple.
For me, more caution about using obsolete version is good enough. (still I prefer version up ruby.)
But I did not drop ruby 1.8.7., so I want to wait for other ffaker people's approval.

rstacruz:
> ## Ruby 1.8.7
> In the case you're still using Ruby 1.8.7, you cannot use the last version of ffaker:

Grammar nitpick: the first clause in that sentence is redundant (the heading is already Ruby 1.8.7). Here's my suggested wording:

> ## Ruby 1.8 support
> FFaker 1.31.0 is the last version that works with legacy Ruby versions. This version is no longer supported; recent releases will only target supported Ruby versions.
 
> ```rb
gem 'ffaker', '1.31.0'
```

> Finally, you'll have to use the `Faker` class (v1) instead of `FFaker` (v2 and later).

tejasbubane:
> recent releases will only target supported Ruby versions.

You might want to specify what is `supported`. Is it officially supported MRI versions or the ones we have in `travis.yml`?

zedtux:
PR updated and rebased on master

zedtux:
@tejasbubane and @marocchino can you please let me know if the PR is fine now?

zedtux:
@tejasbubane can you please review my PR?

tejasbubane:
@zedtux I think the section should not be before `why ffaker`. It should be somewhere below. 

Content-wise this looks good enough to me. But I am not a maintainer of the project and do not have the rights to merge - just here to help. Unfortunately you will have to wait for the maintainers to get back on this.

zedtux:
Oh, thank you @tejasbubane for your comment. I should check with @marocchino I guess.

zedtux:
@marocchino can you please give me your input?

zedtux:
No responses, time wasted ... 😩

PR #299, opened 2016-10-09
FFaker::NameTH_EN for transcribed Thai names

firedev:


lib/ffaker/name_th_en.rb
```diff
@@ -0,0 +1,24 @@
+# encoding: utf-8
+
+module FFaker
+  module NameTH_EN
```
marocchino: We dont need underbar here.
See other modules.
https://github.com/ffaker/ffaker/blob/master/lib/ffaker/identification_es_cl.rb#L6

test/test_name_th_en.rb
```diff
@@ -0,0 +1,34 @@
+# encoding: utf-8
+
+require 'helper'
+
+class TestFakerNameTH_EN < Test::Unit::TestCase
```
marocchino: Same above.

firedev:
Here you go, wish I didn't have to re-define methods from the upper module, but I guess that's how it works.

marocchino:
Thanks

PR #311, opened 2016-12-29
Feature/lorem ru russian text

mechos3d:
Hi there. Thanks for your work !

We needed our tests to use russian letters, so I added this to ffaker.
It was made on the basis of LoremUA (ukranian) just by changing words and available letters.
Words are mostly scientific and general terms, no rude or offensive stuff.


marocchino:
Thanks!

PR #314, opened 2017-01-15
Add japanese address

kojino:
I didn't find Japanese address FFaker in the list, so I implemented FFaker::AddressJA and its test.
This FFaker generates postal codes, streets, cities, etc specific to the Japanese address format.

Note that there are three main address formats in Japan: one for special destricts in Tokyo, one for large rural cities, and one for the rest.


kojino:
The CI tests are still failing, but this is due to test_name_ar.rb and not Address::JA files.

test/test_address_ja.rb
```diff
@@ -0,0 +1,58 @@
...
+  def test_city
+    assert_match(japanese_regex('市'), FFaker::AddressJA.city)
+  end
+
+  def test_county
+    assert_match(japanese_regex('郡'), FFaker::AddressJA.county)
+  end
+
+  def test_prefecture
+    assert_match(japanese_regex('都|道|府|県'), FFaker::AddressJA.prefecture)
```
marocchino: これは（xxx都|道|府|県）で解析されますね。

kojino: '[都道府県]'　に変更しておきました！

marocchino:
Thanks!

test/test_address_ja.rb
```diff
@@ -0,0 +1,58 @@
...
+    assert_match(japanese_regex('村'), FFaker::AddressJA.village)
+  end
+
+  def test_designated_city
+    assert_match(japanese_regex('市'), FFaker::AddressJA.city)
+  end
+
+  def test_city
+    assert_match(japanese_regex('市'), FFaker::AddressJA.city)
+  end
```
AlexWayfer: I'm not an expert in Japanese, so I don't know how to fix it in more correct way, but not every entry from `lib/ffaker/data/address_ja/city` (lines 14, 20, 23–25, etc.) or even `lib/ffaker/data/address_jp/city` (line 40) has `市`

marocchino: Ah. right. Thanks for your report!

PR #316, opened 2017-02-15
New imei generator

metalelf0:
The included imei code generator is flawed, as it generates codes that aren't valid IMEIs. You can double check it with tools like [this online validator](http://imei-number.com/imei-validation-check/) or [this gem by gcimmino](https://github.com/gcimmino/rails_imei_validator).

I replaced it with a function that generates valid IMEI codes, and removed a not-necessary check on deterministic output for the generation with a provided serial code. It's possible to create different valid IMEI codes with the same serial number, so there's no need to make it deterministic.

### DIFF
--- a/lib/ffaker/phone_number.rb
+++ b/lib/ffaker/phone_number.rb
@@ -56,11 +56,17 @@ module FFaker
       # AA-BBBBBB-CCCCCC-D
       characters = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
       rbi_codes  = %w(01 10 30 33 35 44 45 49 50 51 52 53 54 86 91 98 99)
+      serial_number ||= rand(1_000_000)
+      serial_number = sprintf("%06d", serial_number).chars.map(&:to_i)
       first_two_chars = rbi_codes.sample
       characters[0] = first_two_chars.chars.map(&:to_i)[0]
       characters[1] = first_two_chars.chars.map(&:to_i)[1]
-      2.upto(13) do |current_position|
+      # serial number part
+      2.upto(7) do |current_position|
         characters[current_position] = (0..9).to_a.sample
+      end
+      8.upto(13) do |current_position|
+        characters[current_position] = serial_number[current_position - 8]
       end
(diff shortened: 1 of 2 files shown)


metalelf0:
Failing tests in the Travis CI are not related to this PR. They are on master too.

lib/ffaker/phone_number.rb
```diff
@@ -54,23 +54,28 @@ def phone_calling_code
...
-
-      imei_base = tac + serial_number
-
-      check_digit = 0
-      base_digits = imei_base.split('').map(&:to_i)
-      base_digits.each_with_index do |digit, i|
-        check_digit += i.even? ? 2 * digit : digit
+      serial_number = format('%06d', serial_number).chars.map(&:to_i)
+      first_two_chars = rbi_codes.sample
+      characters[0] = first_two_chars.chars.map(&:to_i)[0]
```
marocchino: (maybe)
```
characters[0, 2] = first_two_chars.chars.map(&:to_i)
```

lib/ffaker/phone_number.rb
```diff
@@ -54,23 +54,28 @@ def phone_calling_code
...
-      serial_number = sprintf('%06d', serial_number)
-
-      imei_base = tac + serial_number
-
-      check_digit = 0
-      base_digits = imei_base.split('').map(&:to_i)
-      base_digits.each_with_index do |digit, i|
-        check_digit += i.even? ? 2 * digit : digit
+      serial_number = format('%06d', serial_number).chars.map(&:to_i)
+      first_two_chars = rbi_codes.sample
```
marocchino: (must)
Make this deterministic.
```ruby
first_two_chars = fetch_sample(rbi_codes)
```

lib/ffaker/phone_number.rb
```diff
@@ -54,23 +54,28 @@ def phone_calling_code
...
+      characters[0] = first_two_chars.chars.map(&:to_i)[0]
+      characters[1] = first_two_chars.chars.map(&:to_i)[1]
+      # serial number part
+      2.upto(7) do |current_position|
+        characters[current_position] = (0..9).to_a.sample
```
marocchino: same here.

marocchino:
Thanks!

PR #329, opened 2017-08-02
Add day_of_week to Time

tvarley:
- FFaker::Time.day_of_week   (short day names i.e. Mon, Tue)
- FFaker::Time.day_of_week( long: true )   (Long day names i.e. Monday, Tuesday)


marocchino:
Thanks

PR #330, opened 2017-08-04
Added CompanyIT#partita_iva method to generate valid Italian VAT numbers

aaronsama:
Added a method to generate valid VAT numbers for Italian companies. This should be compatible with validations that check the format of a VAT number.


marocchino:
Thanks

PR #332, opened 2017-09-02
Removed/fixed duplicated code

blocknotes:
- name: removed duplicated method html_safe_last_name
- name_sn: fixed duplicated case
- company_se: fixed duplicated case


marocchino:
Thanks

PR #312, opened 2016-12-31
Added Arabic Names

Moafak:

PR #337, opened 2017-09-14
Fix test_name_ar.rb

aycabta:
The `test_name_ar.rb` is added at https://github.com/ffaker/ffaker/pull/312.

Some names of `FFaker::NameAR::FIRST_NAMES_MALE` and `FFaker::NameAR::LAST_NAMES` have space, but some tests on `test_name_ar.rb` try to separate full name by space, and there are spaces outside between first name and last name. So the tests fail stochastically.

This Pull Request fixes it.


marocchino:
Thanks

PR #339, opened 2017-10-31
Replace mass require with autoload

jhawthorn:
Most users will only want a few modules from ffaker, but we always import all of them. This is a little bit slow and eats some memory (~2MB)

This changes ffaker's modules to load only as they are needed, using only the memory and startup time that is needed. This might be slightly more work to maintain, but only slightly.

An unscientific test on my laptop:

**Before**

Usually about 60 to 90 milliseconds

```
$ time ruby --disable-gems -I lib -rffaker -e 'p FFaker::Lorem.sentence'
"Quia optio eaque ullam ut."
0.07s user 0.02s system 99% cpu 0.091 total
```

**After**

Always around 20 milliseconds, most of which is ruby's startup

```
$ time ruby --disable-gems -I lib -rffaker -e 'p FFaker::Lorem.sentence'
"Molestiae enim harum nihil at qui."
0.02s user 0.00s system 96% cpu 0.021 total
```

```
$ time =ruby --disable-gems -e ''
ruby --disable-gems -e ''  0.01s user 0.00s system 95% cpu 0.014 total
```

So this should save 40 to 70 milliseconds of startup time and up to ~2MB of memory.

EDIT: spec failure on 1.9 unrelated and pre-existing. See #340


marocchino:
make sence :)

PR #346, opened 2018-01-09
Updated password generation

makketagg:
Changed `regexp` in `FFaker::Internet::password`. Works with only numbers, letters and `_`


marocchino:
Thanks!

PR #349, opened 2018-01-29
Add orly book cover generator

artplan1:
I found this cool book covers generator: https://dev.to/rly

It's cool, but I see problems with it:
1. It's hosted on heroku (probably free)
2. Author probably should be notified about usage
3. It doesn't return url with extension

I'm ready to add more tests if this generator is useful in gem

<details>
<summary>Click to see sample image</summary>

</details>

### DIFF
--- a/lib/ffaker/book.rb
+++ b/lib/ffaker/book.rb
@@ -2,6 +2,8 @@
 
 module FFaker
   module Book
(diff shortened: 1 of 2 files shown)


marocchino:
Thanks!

PR #351, opened 2018-02-01
Improve FFaker::Time.between docs and implementation

eikes:
The FFaker::Time.between method worked only for parameters of the Time class. Now it works with Date, DateTime, ActiveSupport::TimeWithZone or String objects.


marocchino:
Thanks!

PR #356, opened 2018-04-05
Create FFaker image module

nicolas-brousse:
Create a `FFaker::Image` module to get a file url or an image file from https://placeholder.com/.  

Useful to have generated images file on factories for tests.

Usage example with ActiveStorage:

```ruby
FactoryBot.define do
  factory :post do
    title { FFaker::Lorem.words(2).join(" ") }
    body  { FFaker::Lorem.paragraphs.join("\n\n") }

    after :build do |post|
      post.image.attach(
        io: FFaker::Image.file("1280x720"),
        filename: "#{FFaker::Lorem.word}.jpg",
        content_type: "image/jpg",
      )
    end
  end
end
```


lib/ffaker/image.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def file(size = '300x300', format = 'png', bg_color = :random, text_color = :random, text = nil)
+      download_file(url(size, format, bg_color, text_color, text))
+    end
+
+    private
+
+    def check_size!(size)
+      return true if size =~ /^[0-9]+x[0-9]+$/
```
marocchino: "1x1\nsomestring" will pass this. we can use `\A\d+x\d+\z` instead of it.

marocchino:
@nicolas-brousse Sorry for late response. 😅
You can fix it or just leave a comment. I will merge this after that.

lib/ffaker/image.rb
nicolas-brousse: `return true if size =~ /\A\d+x\d+\z/` like this?

marocchino: Yes. 😅

nicolas-brousse:
@marocchino I did the change 😃

lucascaton:
@marocchino when will this be available?
Could you bump the gem version, please?

marocchino:
@lucascaton I released 2.10.0 just now.

lucascaton:
Thank you! 🙂

PR #371, opened 2018-09-02
India locale names & city

alagu:
-  Adding Indian locale names - Male & Female
- Adding Indian Cities from Wikipedia


alagu:
Thanks for merging upstream, @marocchino!

PR #374, opened 2018-10-01
Add Japanese genders.

walerian777:
Added a new `FFaker::GenderJP` module including a `random` (aliases: `maybe`, `sample`) method, which returns a gender in Japanese.


D-system:
For Japanese it's `JA` not `JP`.

PR #391, opened 2019-04-05
Update Japanese language code for Gender

D-system:
JP is the country code.
JA is the language code.

Introduce by https://github.com/ffaker/ffaker/pull/374

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,7 +6,7 @@
   - Add NameTW, PhoneNumberTW [@pominx]
   - Add AnimalCN, EducationCN [@ckyOL]
   - Add NameIN [@alagu]
-  - Add AnimalPL, GenderPL, GenderJP, PhoneNumberPL [@walerian777]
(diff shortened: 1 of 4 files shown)


D-system:
Just realize even the merge was 2 days ago, it has already been released. Might not be great to change it.

marocchino:
Sorry it is my bad. This is a somewhat complicated issue.

Like an address, apparently belonging to a country, uses a country code.
So if the country code came first, gender might have been added as a country code.
GenderKR is that case.
 
I thought that would be the case in GenderJP as well. but it was JA.

in this repository, suffix could be:

1. country code
2. language code
3. or both

I would like to solve this problem in the next major release, but I think it would be better to align it with JA now.

marocchino:
On the semver side, however, it is not advisable that the released module disappear, so it would be better to move the logic to GenderJA and modify GenderJP to use GenderJA.

marocchino:
close with 20602520c04ce67da2c3c0513b72f9f05b85ae14

PR #393, opened 2019-04-08
Changing mobile_phone_prefix to reflect current prefixes

Kutomore:
For a while now brazillian mobile phone numbers have had the fixed 9 instead of only in some special cases, this change reflects that.


Kutomore:
@marocchino Hi, apparently there was an unrelated error on the CI? Could you rerun it please?

Volosh1n:
I experienced the same test failing sometimes while running tests on my PC locally. Seems like that test should be fixed by someone who knows japanese. :eyes:

PR #394, opened 2019-04-08
Drop support of under 2.3

marocchino:
https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/


marocchino:
@ffaker/team anyone confirm this?

ffaker.gemspec
```diff
@@ -7,7 +7,7 @@ Gem::Specification.new do |s|
   s.version           = '2.11.0'
   s.date              = '2018-01-04'
```
marocchino: Oops, I forgot update dates.

PR #403, opened 2019-06-28
Global refactoring with Rubocop

Volosh1n:
_This is a large PR and I want and should do a big description._

I found out so many rubocop warnings while doing #402 . So, I think it's time to change things a little.

First of all, I disabled some cops which seem excessive. I'll explain this decision by describing every disable.
1. `Style/ModuleFunction`
Actually, EVERY module has this issue because of `extend self`.
It is not big deal to keep `extend self` IMO. Making changes in modules' logic just to satisfy this cop could be meaningless.
2. `Style/Documentation`
Not all modules have documentation comments. Someone write their code with them, someone not. FFaker have both documented and not documented modules. So, this is why we can skip this cop for existing and future modules.
3. `Style/FrozenStringLiteralComment`
Seems like some of the tests depends on freezed or not freezed strings. If we keep this cop enabled and add this comment via `rubocop -a` run to each file. we can face the crashing tests, while their modules actually working correctly. Anyway, I think we can just keep existing frozen string literal comments and not demand it from contributors in future modules. Not the most elegant move, but it's just my suggestions to consider.

Most of other changes in this PR includes:
- keeping modules DRY;
- redundant whitespaces remove;
- adding guard clause in many `case rand(0..1) ...` or simple `if...else` conditions. Adding empty line after existing guard clauses;
- linebreaks to fit into max 100 characters in each line;
- reducing lines in methods to fit 10 max;
- reducing complex logic while keeping same method result (to satisfy `Metrics/AbcSize` cop);
- freezing mutable constants;
- changing naming of methods or constants with incorrect case;
- etc.

Also, some modules (`FFaker::NatoAlphabet`, `FFaker::NameCS`) were bloated by data (storing it in large arrays). This modules was cleaned by excluding data to `data` folder to keep modules clear.

So, this is my suggestions on making FFaker code a bit cleaner. Hope it seems good.


lib/ffaker/company_it.rb
```diff
@@ -33,7 +33,9 @@ def partita_iva
       base = matricola + office_code.chars.map(&:to_i)
 
       # Thanks https://stackoverflow.com/a/9189731/1627766
-      sum = base.reverse.each_slice(2).flat_map { |x, y| [(x * 2).divmod(10), y || 0] }.flatten.reduce(:+)
+      sum = base.reverse.each_slice(2).flat_map do |x, y|
+        [(x * 2).divmod(10), y || 0]
+      end.flatten.reduce(:+)
```
marocchino: howw about `.flat_map { |x, y=0| (x * 2).divmod(10) + y }.sum`?

lib/ffaker/identification_es_cl.rb
```diff
@@ -21,7 +22,9 @@ def rut
     private
 
     def dv(rut)
-      total = rut.to_s.rjust(8, '0').split(//).zip(%w[3 2 7 6 5 4 3 2]).collect { |a, b| a.to_i * b.to_i }.inject(:+)
+      total = rut.to_s.rjust(8, '0').split(//).zip(%w[3 2 7 6 5 4 3 2]).collect do |a, b|
```
marocchino: I think `.sum { |a, b| a.to_i * b.to_i }` is better here.
http://ruby-doc.org/core-2.6.3/Array.html#method-i-sum

Just for curious, Why you expand every {}  in the middle of changing? Is that better readable to you?

lib/ffaker/lorem.rb
```diff
@@ -7,7 +7,8 @@ module Lorem
     extend self
 
     def characters(character_count = 255)
-      Array.new(character_count < 0 ? 0 : character_count).map { fetch_sample(CHARACTERS) }.join
+      character_count = 0 if character_count.negative?
+      Array.new(character_count).map { fetch_sample(CHARACTERS) }.join
```
marocchino:
```suggestion
      fetch_sample(CHARACTERS, count: character_count).join
```

lib/ffaker/phone_number_sn.rb
```diff
@@ -64,15 +64,9 @@ def homework_number
...
-      case rand(0..9)
-      when 0 then homework_number
-      when 9 then homework_number
-      when 3 then mobile_number
-      when 5 then mobile_number
-      when 6 then mobile_number
-      when 8 then mobile_number
-      else        homework_number
-      end
+      return homework_number if rand(0..1).zero?
```
marocchino: is not about this pr but, ArrayUtils.rand is deprecated.
```suggestion
      return homework_number if fetch_sample(0..1).zero?
```

lib/ffaker/identification_es_cl.rb
Volosh1n: I just wanted to reduce line length to fit into 100 symbols max :smile:

lib/ffaker/lorem.rb
Volosh1n: This change will use only 26 letters, even if you set `character_count` bigger than 26. Sad.

Volosh1n:
I added code from your suggestions, thank you for your help.

PR #410, opened 2019-08-30
Bank module improvements

Volosh1n:
Hi! :wave: 
Here are some suggestions to improve Bank module:

- Using `FFaker.numerify` in card number generator instead of finite number of card numbers.
- Card expiry date generator was just a duplicate of `FFaker::Time.date` without actual card expiry context. Made it look like `mm/yy` format.
- Guard claused `check_country_existence` method.
- Added these new methods to `REFERENCE.md`.

Hope it makes module better.


lib/ffaker/bank.rb
```diff
@@ -92,13 +92,11 @@ def iban(country_code: nil)
...
     def card_number
-      fetch_sample(CARD_NUMBERS)
+      FFaker.numerify('#### #### #### ####')
     end
 
-    def card_expiry_date(**params)
-      params[:year_range] ||= 5
-      params[:year_latest] ||= -params[:year_range] # for the future
-      FFaker::Time.date(params)
+    def card_expiry_date(params = { year_range: 5, year_latest: -5 })
```
marocchino: (maybe) This will cover 'mm/yyyy' form. 
```suggestion
    def card_expiry_date(year_range: 5, year_latest: -5, format: '%m/%y')
      FFaker::Time.date(params).strftime(format)
```

Volosh1n: done

PR #414, opened 2019-09-06
Remove unexpected char in email

### LINKED ISSUE #399: FFaker::Internet email, safe_email and free_email generating invalid email addresses
In FFaker::Internet methods email, safe_email and free_email are generating email addresses with underscores which are not valid.

marocchino:
close #399

### DIFF
--- a/lib/ffaker/internet.rb
+++ b/lib/ffaker/internet.rb
@@ -39,7 +39,7 @@ module FFaker
 
       return sanitize(Name.first_name) if rand(0..1).zero?
 
-      [Name.first_name, Name.last_name].join(fetch_sample(%w[. _])).downcase
+      [sanitize(Name.first_name), sanitize(Name.last_name)].join(fetch_sample(%w[. _]))
     end
 
     def domain_name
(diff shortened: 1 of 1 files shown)


davidhrbac:
https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address

lib/ffaker/internet.rb
```diff
@@ -39,17 +39,15 @@ def user_name(name = nil)
...
 
     def domain_name
       [domain_word, domain_suffix].join('.')
     end
 
     def domain_word
-      dw = Company.name.split(' ').first
-      dw.gsub!(/\W/, '')
-      dw.downcase
+      sanitize(Company.name)
```
Volosh1n: I'd say this method be better returning the same value
```ruby
sanitize(Company.name.split(' ').first)
```

Volosh1n:
```suggestion
sanitize(Company.name.split(' ').first)
```

Volosh1n: I'd say this method should return same value, but sanitized.
```ruby
sanitize(Company.name.split(' ').first)
```

marocchino: 5e4f8dc

PR #418, opened 2019-10-01
Add Polish colours.

walerian777:
Added a new `FFaker::ColorPL` module including a `#name` method which returns Polish colour names.


Volosh1n:
Thank you! Also, I'd suggest to add some examples to `REFERENCE.md` to make people know about your module. What do you think?

walerian777:
@Volosh1n Of course - it's already updated.

PR #419, opened 2019-10-02
Add Polish identification.

walerian777:
Added a new `FFaker::IdentificationPL` module for generating Polish identification numbers.
The following methods are available:
* `pesel` - generates a [PESEL](https://en.wikipedia.org/wiki/PESEL) - Polish national identification number
* `identity_card` - generates a [Polish identity card](https://en.wikipedia.org/wiki/Polish_identity_card) number
* `drivers_license` - generates a [Polish driver's licence](https://en.wikipedia.org/wiki/Driving_licence_in_Poland) number

### DIFF
--- a/lib/ffaker/identification_pl.rb
+++ b/lib/ffaker/identification_pl.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'date'
+
+module FFaker
+  module IdentificationPL
+    extend ModuleUtils
+    extend self
+
+    # Polish national identification number
+    # https://en.wikipedia.org/wiki/PESEL
+    def pesel
+      date = generate_valid_pesel_date
+      birthdate = pesel_birthdate(date)
+      serial_number = FFaker.numerify('####')
+      checksum = pesel_checksum(birthdate, serial_number)
+      "#{birthdate}#{serial_number}#{checksum}"
+    end
+
+    # Polish identity card number
+    # https://en.wikipedia.org/wiki/Polish_identity_card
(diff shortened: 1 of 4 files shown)


lib/ffaker/identification_pl.rb
```diff
@@ -0,0 +1,77 @@
...
+    def drivers_license
+      FFaker.numerify('#####/##/####')
+    end
+
+    private
+
+    def generate_valid_pesel_date
+      from = Date.new(1800, 1, 1)
+      to = [Date.today, Date.new(2299, 12, 31)].min
+      fetch_sample([*from..to])
```
marocchino: hmm... is this casting needed?
```suggestion
      fetch_sample(from..to)
```

walerian777: Unfortunately, it's needed - `fetch_sample` performs on its argument, so after the suggested change we would receive:
`NoMethodError: undefined method `sample' for #<Range:..>`

We could probably make `fetch_sample` work with ranges, but that's probably worth a separate PR.

marocchino: oh, I see. `fetch_sample` only receiving array. 
But still It is too big array for just one date.

Volosh1n: I think `FFaker::Time` module could be used. For example, `FFaker::Time#between` method.

marocchino: OK, just `fetch_sample(from, to)` will be fine after #420 merged.

PR #433, opened 2020-03-12
Making ability to set gem version in only one place of project

Volosh1n:
Also, removed `require 'set'` since it's a part of Ruby.

### DIFF
--- a/ffaker.gemspec
+++ b/ffaker.gemspec
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
+lib = File.expand_path('lib', __dir__)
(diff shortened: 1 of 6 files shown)


ffaker.gemspec
```diff
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'version'
```
AlexWayfer, line 5: It's bad and old practice, please use `require_relative`.

PR #441, opened 2020-05-07
# US Country name & code updated.

nigelgomesot:
Recently when using `US` Country FFaker, identified that the `country` & `country_code` values were random. submitting this improvement. Please feel free to post a feedback.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -817,6 +817,8 @@
 | `street_name` | Mike Plain, Wisozk Ports, Zaida Causeway |
 | `street_suffix` | Mall, Harbor, Track |
 | `time_zone` | Pacific/Midway, Europe/Bratislava, Asia/Muscat |
+| `country` | United States |
+| `country_code` | US |
 | `uk_country` | ❗ *[uk_country] is deprecated. For UK addresses please use the AddressUK module* |
 | `uk_county` | ❗ *[uk_county] is deprecated. For UK addresses please use the AddressUK module* |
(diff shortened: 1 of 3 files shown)


Volosh1n:
Hi, @nigelgomesot !
Thanks for your PR, but I think there is missunderstanding about AddressUS module. AddressUS just generates random data in English. It is not about show something about USA exactly (like, the name of the USA country). So, `country` should just generate random countries, but in English language.

nigelgomesot:
Hey @Volosh1n  thanks for your feedback!
understood now, my interpretation of the AddressUS was precise Address in US (including country), eg: similar to [Indian Address](https://github.com/ffaker/ffaker/blob/8e25825bf05d75266f2e888a8daf13ceb30d9b09/lib/ffaker/address_in.rb#L46).
But after looking at other `address_**.rb` files it seems we have not specified `country` for them explicitly. 
Let me know if this change is irrelevant, will close it.

Volosh1n:
I think we should keep `AddressUS.country` to still generate random countries (for those projects, who already use this method to generate random countries in English). Didn't know about that method in Indian Address, to be honest. Maybe @marocchino has another point for this.
Thanks for your attention to gem anyway!

marocchino:
I agree that it is not consistent and confusing.
In the next major version, I think we need to fix the confusion of mixing country codes and language codes, but for now I think it would be better to just close this.

It was a very good point. Thank you anyway.

nigelgomesdev:
@marocchino thanks for the clarification 👍

PR #442, opened 2020-05-13
Make sure dutch zipcodes never start with a 0

gerard76:
I had a flipper spec that failed because the dutch zipcode did not always pass the mustar. A dutch  zipcode never starts with a 0.

For reference see the map at https://en.wikipedia.org/wiki/Postal_codes_in_the_Netherlands

### DIFF
--- a/lib/ffaker/address_nl.rb
+++ b/lib/ffaker/address_nl.rb
@@ -12,7 +12,7 @@ module FFaker
     POSTAL_CODE_FORMAT = ['#### ??'].freeze
 
     def postal_code
-      FFaker.bothify(POSTAL_CODE_FORMAT).tap { |p| p[0] = rand(1..9).to_s }
+      FFaker::String.from_regexp(/[1-9]\d{3} [A-Z]{2}/)
     end
 
     def zip_code
--- a/lib/ffaker/address_nl.rb
+++ b/lib/ffaker/address_nl.rb
@@ -12,7 +12,7 @@ module FFaker
     POSTAL_CODE_FORMAT = ['#### ??'].freeze
 
     def postal_code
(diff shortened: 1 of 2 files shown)


lib/ffaker/address_nl.rb
```diff
@@ -12,7 +12,7 @@ module AddressNL
     POSTAL_CODE_FORMAT = ['#### ??'].freeze
 
     def postal_code
-      FFaker.bothify POSTAL_CODE_FORMAT
+      FFaker.bothify(POSTAL_CODE_FORMAT).tap { |p| p[0] = rand(1..9).to_s }
```
marocchino, line 15: I want to avoid string mutation.

(maybe)
```
    POSTAL_CODE_FORMAT = ['### ??'].freeze

    def postal_code
      "#{rand(1..9)}#{FFaker.bothify(POSTAL_CODE_FORMAT)}"
```

marocchino: or 
```rb
FFaker::String.from_regex(/[1-9]\d{3} [a-z]{2}/)
```

test/test_address_nl.rb
```diff
@@ -23,7 +23,7 @@ def test_province
   end
 
   def test_zip_code
-    assert_match(/^\d{4}\s\w{2}$/, @tester.zip_code)
+    assert_match(/^[1-9]\d{3}\s\w{2}$/, @tester.zip_code)
```
marocchino, line 26:
```suggestion
    assert_match(/^[1-9]\d{3} [A-Z]{2}$/, @tester.zip_code)
```

marocchino:
💚

PR #444, opened 2020-05-23
2.15.0 update

Volosh1n:
Hi! There are some new modules added. Could it be a new release? :sweat_smile:


marocchino:
If there is an update, I tried to do it once a month, but I forget it every time. 😅  Thanks for this work.

PR #445, opened 2020-06-07
Enable Rubocop's pending cops

Volosh1n:
Enabled pending cops after [this](https://github.com/rubocop-hq/rubocop/issues/7850) Rubocop's update. From now, all new cops will be automatically enabled without all those warnings on each rubocop run.

### DIFF
--- a/lib/ffaker/string.rb
+++ b/lib/ffaker/string.rb
@@ -50,7 +50,7 @@ module FFaker
       token = tokens.shift
 
       case token
-      when '?' then
+      when '?'
         # TODO: Let ? generate nothing
         '' # We already printed its target
       when '+' then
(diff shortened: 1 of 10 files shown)


lib/ffaker/string.rb
```diff
@@ -50,7 +50,7 @@ def process_token(tokens)
       token = tokens.shift
 
       case token
-      when '?' then
+      when '?'
         # TODO: Let ? generate nothing
         '' # We already printed its target
       when '+' then
```
marocchino, line 56: 🧐 I thought this `then` also unnessasary.

Volosh1n: @marocchino, just Rubocop's autocorrection :sweat_smile:

PR #451, opened 2020-08-01
Apply rubocop autofix

marocchino:


marocchino:
It looks okay, but I'd like you to cross check.

PR #468, opened 2021-01-29
Test with github action & ruby 3

marocchino:


lib/ffaker/image.rb
```diff
@@ -42,7 +42,7 @@ def check_format!(format)
     def download_file(url)
       file = Tempfile.new('ffaker_image')
       file.binmode
-      file << Kernel.open(url).read
+      file << URI.open(url).read
```
marocchino, line 45: https://bugs.ruby-lang.org/issues/15893

rstacruz:
Hey mate... All these changes are looking great! :)

PR #271, opened 2016-04-17
Add "Freedom Ipsum" text generator option

lynnd335:
Add "[Freedom Ipsum](http://ameripsum.win/)" ultra-patriotic filler text to FFaker ... reference MD file updated accordingly.

PR #469, opened 2021-02-09
Add "Freedom Ipsum" text generator option

hmaack:
In order to help #271 get merged into `main` I fixed the last issues. 

cc @lynnd335


hmaack:
Hi @AlexWayfer because you probably reviewed recently #271 I would like to get a review. Because I helped this guy getting his PR merged into main. 🤣

lib/ffaker/freedom_ipsum.rb
```diff
@@ -0,0 +1,20 @@
...
+    extend ModuleUtils
+    extend self
+
+    def word
+      fetch_sample(FREEDOM_WORDS)
+    end
+
+    def words(num = 3)
+      fetch_sample(FREEDOM_WORDS, count: num)
+    end
```
AlexWayfer, line 18: Where is definition of this constant?

I don't see a search in files on GitHub for specific branch. Search in changes and `main` branch gives nothing.

Also I don't understand why tests pass if there is error, it should be changed too then.

lib/ffaker/data/freedom_ipsum/freedom_words
```diff
@@ -0,0 +1,161 @@
...
+online multiplayer
+controversial
+the media
+velcro
+foreign policy
+national security
+elections
+Home Depot
+ESPN2
+weekends
```
AlexWayfer, line 161: I don't like files without newline at the end.

https://unix.stackexchange.com/a/18789/394247
https://thoughtbot.com/blog/no-newline-at-end-of-file

test/test_freedom_ipsum.rb
```diff
@@ -0,0 +1,21 @@
...
+require 'helper'
+class TestFreedomIpsum < Test::Unit::TestCase
+  include DeterministicHelper
+
+  assert_methods_are_deterministic(
+    FFaker::FreedomIpsum,
+    :word, :words
+  )
+
+  TEST_REGEX = /\+1|[ a-z]+/i.freeze
```
AlexWayfer, line 12: 1. I don't see `+1` text, I don't know why it's here.
2. But I see strings like `7-Eleven` and `1776`, so there should be something like `/[a-z0-9 -]+/i`

AlexWayfer (commented):
It's unexpected for me, but OK, here is my review, I hope it's not too bad.

lib/ffaker/freedom_ipsum.rb
hmaack: Hi @AlexWayfer. This is the magic part in this gem. It's a long time ago that I contributed for this gem. But I guess you are looking for this monkey patch `const_missing` https://github.com/ffaker/ffaker/blob/9ed6e8c7d51fef88f94110f7701fc03e0390fda5/lib/ffaker/utils/module_utils.rb#L15 

In case the any const is missing it's looking into the data path to resolve this selection.

test/test_freedom_ipsum.rb
hmaack: Good catch. Didn't checked this.

lib/ffaker/freedom_ipsum.rb
AlexWayfer: Give me a few days please, I want to investigate it and maybe try to remake: such "magic" is not good, obvious and fast.

hmaack: I appreciate your refactoring motivation and I completely agree. You will find this behaviour in almost every module in this gem. So its basically about this `const_missing` magic. The ugly part to use ruby in my opinion, hard to maintain and easy to create bugs. That was the ruby style 2013. 🤣 But I would recommend to create a new issue / PR for this refactoring, but take your time. 👍

marocchino (approved):

lib/ffaker/freedom_ipsum.rb
AlexWayfer: Nevermind: I saw also #470, I was trying to improve this too, and faced with inconsistency between `lorem` in different languages… I don't have a wish to refactor these things anymore, at least for now.

Issue #424, opened 2020-02-22
TestLoremFR tests fail randomly


dleidert:
Hi there,

we observe some random test failures regarding the tests in `test/test_lorem_fr.rb`. It seems the regexes used for sentence*, paragraph* and phrase* tests don't apply if the result starts with a capital accented letter. This simple patch https://salsa.debian.org/ruby-team/ruby-ffaker/blob/master/debian/patches/0001-Fix-TestLoremFR-tests.patch seems to work.

I also observe random test failures with TestAddressJA's test_city() test:
<https://salsa.debian.org/ruby-team/ruby-ffaker/-/jobs/574659#L2846>
<https://buildd.debian.org/status/fetch.php?pkg=ruby-ffaker&arch=all&ver=2.8.1-2&stamp=1582390001&raw=0>

Volosh1n:
Hi, @dleidert !
Thank you for information. Could you, please, open PR for that LoremFR fix? Also, `AddressJA.city` seems to be fixed in #411

hmaack:
Looks like @marocchino merged a fix. Feel free to close this one @marocchino.

PR #470, opened 2021-02-12
fix: change the regex to cover all scenarios in lorem_fr

abnersajr:
This PR aims to close the issue #424.

Taking the suggestion from the description, I've moved all regexes to constants, changing in all test cases.


marocchino:
Thanks! <3

PR #482, opened 2021-04-25
Add games

nicholaskillin:
Hey!

I wanted to generate fake Board Game data. I thought that it might be nice to name it `Games` so that it could be easily expanded later for other types of games if we wanted to.

I'm open to thoughts or changes, so let me know what you think!


nicholaskillin:
Sorry about that! Copied a test for another module and forgot to rename the class. 

I corrected that and ran the entire test suite locally with no errors this time.

Volosh1n (approved):

test/test_game.rb
```diff
@@ -0,0 +1,28 @@
...
+    FFaker::Game,
+    :title, :category, :mechanic
+  )
+
+  def setup
+    @tester = FFaker::Game
+  end
+
+  def test_title
+    assert_match(/[\sa-z]+/, @tester.title)
```
marocchino, line 18: Note: we need to clean up non-all-range-cover regexp all at once later.

marocchino:
👍🏽

test/test_game.rb
nicholaskillin: @marocchino I'm working on another PR for a new class. Is there a pattern that you would say is a better example for testing that a method returns a string than the one I followed here?

marocchino: @nicholaskillin 
There are a couple of patterns that aren't very cool right now. 

1. Include
It can be used when a variable is created directly from a constant.
https://github.com/ffaker/ffaker/blob/9ed6e8c7d51fef88f94110f7701fc03e0390fda5/test/test_address_id.rb#L26

2. Regular expression including `\A`, `\z`
Honestly, finding characters in files is very tiring, so I don't want to do it very much.
https://github.com/ffaker/ffaker/blob/9ed6e8c7d51fef88f94110f7701fc03e0390fda5/test/test_address_id.rb#L36 

I'd like to make something like this later.
```rb
assert_characters_in @tester.method, %w[filename1 filename2]

# or

assert_characters_in @tester.method, [@tester::CONST1, @tester::CONST2]
```

PR #496, opened 2021-06-19
Fix gem autoloading

SteveRedka:
`autoload` doesn't seem to be able to require files relative to its directory, unless this directory is required as well. Addition of gem's root directory to $LOAD_PATH seems to fix it.

Also, I added an instruction how to run all tests at once, since there seems to be a lot of red tests.


Volosh1n:
Hello!
To run a single file, you can try something like this:
```bash
rake test TEST=test/test_lorem_br.rb
```
To tun all of the tests, you can try just `rake` or `rake test`

SteveRedka:
Ok, that works

PR #505, opened 2021-08-19
Add Number module

Volosh1n:
I was curious checking the @ngan idea in #503 and decided to add the basic solution.
Hope it fits well

### DIFF
--- a/lib/ffaker/number.rb
+++ b/lib/ffaker/number.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module FFaker
+  module Number
+    extend ModuleUtils
+    extend self
+
+    def number(digits: 1)
+      FFaker.numerify(generate_pattern(digits)).to_i
+    end
+
+    def decimal(whole_digits: 1, fractional_digits: 1)
+      whole_part_pattern = generate_pattern(whole_digits)
+      fractional_part_pattern = generate_pattern(fractional_digits).reverse
+      FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
+    end
+
+    private
+
+    def generate_pattern(digits)
+      raise ArgumentError.new('Digits cannot be less than 1') if digits < 1
+
+      "#{fetch_sample(1..9)}#{"#" * (digits - 1)}"
+    end
+  end
+end
--- a/test/test_number.rb
+++ b/test/test_number.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+class TestNumber < Test::Unit::TestCase
+  include DeterministicHelper
+
+  assert_methods_are_deterministic(FFaker::Number, :number, :decimal)
+
+  def setup
+    @tester = FFaker::Number
+  end
+
+  def test_number
+    assert @tester.number.is_a?(Integer)
+    assert @tester.number.digits.count == 1
+    assert @tester.number(digits: 3).digits.count == 3
+    assert_match(/\d/, @tester.number.to_s)
+    assert_match(/\d{3}/, @tester.number(digits: 3).to_s)
+  end
+
+  def test_number_when_invalid_argument
+    assert_raise(ArgumentError.new('Digits cannot be less than 1')) { @tester.number(digits: 0) }
+  end
+
+  def test_decimal
+    assert @tester.decimal.is_a?(Float)
+
+    whole_digits = rand(2..5)
+    fractional_digits = rand(2..5)
+    decimal = @tester.decimal(whole_digits: whole_digits, fractional_digits: fractional_digits)
+    assert decimal.to_s.split('.').first.size == whole_digits
+    assert decimal.to_s.split('.').last.size == fractional_digits
+  end
+
+  def test_decimal_when_invalid_argument
+    assert_raise(ArgumentError.new('Digits cannot be less than 1')) do
+      @tester.decimal(whole_digits: 0)
+    end
+
+    assert_raise(ArgumentError.new('Digits cannot be less than 1')) do
+      @tester.decimal(fractional_digits: 0)
+    end
+  end
+end
--- a/lib/ffaker.rb
+++ b/lib/ffaker.rb
@@ -173,6 +173,7 @@ module FFaker
   autoload :NameUA, 'ffaker/name_ua'
   autoload :NameVN, 'ffaker/name_vn'
   autoload :NatoAlphabet, 'ffaker/nato_alphabet'
+  autoload :Number, 'ffaker/number'
   autoload :PhoneNumber, 'ffaker/phone_number'
   autoload :PhoneNumberAU, 'ffaker/phone_number_au'
   autoload :PhoneNumberBR, 'ffaker/phone_number_br'


lib/ffaker/number.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module FFaker
+  module Number
+    extend ModuleUtils
+    extend self
+
+    def number(digits: 1)
+      FFaker.numerify(generate_pattern(digits)).to_i
+    end
```
AlexWayfer, line 10: Would we cover a non-zero-at-first case here?

lib/ffaker/number.rb
```diff
@@ -0,0 +1,24 @@
...
+      left_digits_pattern = generate_pattern(left_digits)
+      right_digits_pattern = generate_pattern(right_digits)
+      "#{FFaker.numerify(left_digits_pattern)}.#{FFaker.numerify(right_digits_pattern)}".to_f
+    end
+
+    private
+
+    def generate_pattern(digits)
+      Array.new(digits) { '#' }.join
+    end
```
AlexWayfer, line 22: Why inside curly braces instead of parentheses?

```ruby
pry: main > Array.new(10) { '#' }.join
# => "##########"
pry: main > Array.new(10, '#').join
# => "##########"
```

Volosh1n: Wow. New day - new fact

lib/ffaker/number.rb
Volosh1n: You're right, I'll fix

Volosh1n: Oops. sorry, re-requested review too soon

Volosh1n: Well, seems like now it covered, thanks for this catch

test/test_number.rb
```diff
@@ -0,0 +1,40 @@
...
+  def test_number
+    assert @tester.number.is_a?(Integer)
+    assert @tester.number.digits.count == 1
+    assert @tester.number(digits: 3).digits.count == 3
+    assert_match(/\d/, @tester.number.to_s)
+    assert_match(/\d{3}/, @tester.number(digits: 3).to_s)
+  end
+
+  def test_number_when_invalid_argument
+    assert @tester.number(digits: 0) == nil
```
Volosh1n, line 23: Not sure about it. Maybe return 0 eif digits < 1? Let's discus

lib/ffaker/number.rb
ngan:
```ruby
'#' * digits
```

is even cheaper 😉

test/test_number.rb
ngan: I would actually recommend raising an ArugmentError here. It's simply that, we don't want/understand anything less than 1.

lib/ffaker/number.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      result.tap { |str| str[0] = fetch_sample((1..9).to_a).to_s }.to_i
+    end
+
+    def decimal(left_digits: 1, right_digits: 1)
+      return if left_digits < 1 || right_digits < 1
+
+      left_digits_pattern = generate_pattern(left_digits)
+      right_digits_pattern = generate_pattern(right_digits)
+      "#{FFaker.numerify(left_digits_pattern)}.#{FFaker.numerify(right_digits_pattern)}".to_f
```
ngan, line 22: This needs protections from leading `0` (before the decimal) _and_ trailing `0` (after the zero).

lib/ffaker/number.rb
marocchino: I recommand fetch_sample(1...10**digits)

lib/ffaker/number.rb
marocchino: maybe) number(left_digits + right_digits) / (10**right_digits).to_f

lib/ffaker/number.rb
```diff
@@ -0,0 +1,31 @@
...
+    def number(digits: 1)
+      return if digits < 1
+
+      result = FFaker.numerify(generate_pattern(digits))
+      return result.to_i if result.chars.first != '0'
+
+      result.tap { |str| str[0] = fetch_sample((1..9).to_a).to_s }.to_i
+    end
+
+    def decimal(left_digits: 1, right_digits: 1)
```
ngan, line 17: If we want to get technical, these arguments should be `whole_digits` and `fractional_digits`... leaving it up to you 😄 just pointing it out.

ngan: Better yet, I'm not even sure this really needs to be kwarg. It seems like you'd always want 2 arguments: left and right. So maybe the method should just be `demical(whole_digits, fraction_digits)`

Volosh1n: I think `FFaker::Number.decimal(3, 7)` could be confusing if you don't know what is the first postion argument, what is the second one

lib/ffaker/number.rb
Volosh1n: @ngan @marocchino I've decided to generate whole and fractional with Number#number, since it already has leading 0 protection. Also, removed that private method, which generated # pattern

Volosh1n:
I think it's prettier now. Ready to re-review @marocchino @ngan @AlexWayfer

lib/ffaker/number.rb
ngan: The (fractional) part after the decimal needs trailing 0 protection (not leading).

WHOLE.FRACTIONAL

WHOLE cannot start with a 0
FRACTIONAL cannot end with a 0

Volosh1n: hmm... okay, I'll fix, no need to caps it

ngan: Hehe sorry, it seemed like like the caps is like I'm screaming those words. I meant for it look like (Ruby) constants with all caps. My bad.

Volosh1n: That's ok

Volosh1n: Fixed

lib/ffaker/number.rb
ngan: That's valid. Although if you look api.rubyonrails.org, there's many methods that you just have to red documentation on to know what to pass in as the first, second, third, etc argument. I think verbose is fine too though, but just feels like you should read the docs to know how to use `decimal(..)` anyways. Up to you 😄

lib/ffaker/number.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module FFaker
+  module Number
+    extend ModuleUtils
+    extend self
+
+    def number(digits: 1)
+      raise ArgumentError.new('Digits counter should be 1 or greater') if digits < 1
```
ngan, line 9:
```suggestion
      raise ArgumentError.new('Digits cannot be less than 1') if digits < 1
```

lib/ffaker/number.rb
Volosh1n: @ngan yea, positional arguments are everywhere in ffaker too. Also, if I want some explicit count for fractional part, but I don't mind what will be in whole part, I still have to put some number as first argument, even though it is not necessay for such case. Let's keep kwargs, if other guys don't mind too

lib/ffaker/number.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      pattern = '#' * digits
+      result = FFaker.numerify(pattern)
+      return result.to_i if result.chars.first != '0'
+
+      result.tap { |str| str[0] = fetch_sample((1..9).to_a).to_s }.to_i
```
ngan, line 15: It feels like we're relying on chance to generate a "good" number. How do you feel about not leaving that to chance and always generate a good number on the first pass:
```suggestion
      pattern = '#' * (digits - 1)
      leading_digit = rand(1..9)
      FFaker.numerify("#{leading_digit}#{pattern}").to_i
```

You could extract this out to a private method and allow it to take either leading digit safety or trailing digit safety and reuse the logic for `decimal`.

Volosh1n: Sorry, don't want to sound rude, but what the purpose of such change? I mean.. doesn't it work the same? If it generates "good" number - let it go. if not - replace invalid '0' and that's all.

ngan:
An abstraction could look like this:

```ruby
private

def number(digits: 1)
  ...
  FFaker.numerify(generate_pattern(digits)).to_i
end

def decimal(whole_digits, fractional_digits)
  FFaker.numerify("#{generate_pattern(digits)}.#{generate_pattern(digits).reverse}").to_d
end

def generate_pattern(digits)
  "#{rand(1..9)}#{"#" * (digits - 1)}"
end
```

Volosh1n:
Let the last comment be up to @marocchino too. I think PR is good enough

Volosh1n:
@ngan applied, but replaced rand(1..9) with fetch_sample to make methods pass deterministic tests

lib/ffaker/number.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def number(digits: 1)
+      FFaker.numerify(generate_pattern(digits)).to_i
+    end
+
+    def decimal(whole_part_digits: 1, fractional_part_digits: 1)
+      whole_part_pattern = generate_pattern(whole_part_digits)
+      fractional_part_pattern = generate_pattern(fractional_part_digits).reverse
+      FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
+    end
```
ngan, line 16: What do you think about renaming the argument to:
`whole_part_digits` => `whole_digits`
`fractional_part_digits` => `fractional_digits`

lib/ffaker/number.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    private
+
+    def generate_pattern(digits)
+      raise ArgumentError.new('Digits cannot be less than 1') if digits < 1
+
+      non_zero_numbers = (1..9).to_a
+      "#{fetch_sample(non_zero_numbers)}#{"#" * (digits - 1)}"
```
ngan, line 24: Looks like fetch_sample supports ranges: https://github.com/ffaker/ffaker/blob/b661ddb32027c2abac66a5162b1198e3ea553c3e/lib/ffaker/utils/random_utils.rb#L23

```suggestion
      "#{fetch_sample(1..9)}#{"#" * (digits - 1)}"
```

ngan (approved):
This looks great! Thanks @Volosh1n

ngan:
Looks like this repo has a changelog…should this Pr include update there?

Volosh1n:
@nagan, no, we create additioanl release pr, which includes changelog

marocchino (approved):

technicalpickles:
Could we get a gem release that includes this?

PR #509, opened 2021-10-05
Update the Rakefile to not generate a gemspec

technicalpickles:
It no longer uses jeweler, so shouldn't need to be regenerating the file.
Best practice is to have `git ls-files` in the gemspec itself. It will
be evaluated at the time that the gem is built, and an expanded version
of the gemspec is included in the gem.

This changes the Rakefile to remove the gemspec task, and move the logic
for determining files into the gemspec.

Some other changes related:

- remove check for VERSION file, since no longer using jeweler
- use `gem build` with `--output` to not require mving it after it's built
- add `-Ilib` to `reference` task, otherwise it fails to find ffaker

Closes https://github.com/ffaker/ffaker/issues/190


marocchino:
@technicalpickles
Sorry. I put it on hold because I wasn't familiar with it. I want to merge it, can you rebase it or give the permission to edit it?

technicalpickles:
@marocchino I've rebased, and you should already have access to make edits.

PR #510, opened 2021-10-05
Consistently generate REFERENCE.md

technicalpickles:
Right now, if you generate the file multiple times, you will different
results. This is noteworthy because it is generated as part of
`rake build` and `rake release`.

This is fixed by using assigning a `FFaker::Random.seed`.

Not specifically about a consistent REFERENCE.md, but I've also added
`unique` to the calls to prevent duplicates to get more variety.

While running this, I was getting errors that `Set` wasn't defined, so I added a require to `UniqueUtils` for it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,11 @@
 
   - Add your change HERE
 
+## 2.20.0
+  - Remove date from gem specification [@AlexWayfer]
+  - Add Number module [@Volosh1n]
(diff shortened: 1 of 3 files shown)


marocchino (approved):
LGTM! just need resolve conflict. :)

technicalpickles:
@marocchino done!

PR #512, opened 2021-10-15
Add neighborhood and address complement to AddressBR

az3vedo:
I've added two methods to the AddressBR module:

- Neighborhood: A method to generate brazilian neighborhood names using the most common prefixes

- Complement: A method to generate complement to the addresses, most brazilian e-commerces/stores/information systems use the complement as a secondary address e.g: Apartment 104, Shed

Can you please add the `hacktoberfest-accepted` tag to this PR? It's my first contribution :D

### DIFF
--- a/lib/ffaker/address_br.rb
+++ b/lib/ffaker/address_br.rb
@@ -42,11 +42,16 @@ module FFaker
       FFaker.numerify(fetch_sample(COMPLEMENT))
     end
 
+    def neighborhood
+      prefix = fetch_sample(NEIGHBORHOOD_PREFIXES)
+      "#{prefix} #{NameBR.first_name}"
+    end
+
     def full_address(with_complement: false)
       if with_complement
-        "#{street}, #{building_number}, #{complement}, #{city}, #{state}, Brazil"
+        "#{street}, #{building_number}, #{complement}, #{neighborhood}, #{city}, #{state}, Brazil"
       else
-      "#{street}, #{building_number}, #{city}, #{state}, Brazil"
+      "#{street}, #{building_number}, #{neighborhood}, #{city}, #{state}, Brazil"
       end
     end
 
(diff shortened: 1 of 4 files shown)


az3vedo:
Someone?

test/test_address_br.rb
```diff
@@ -34,4 +34,14 @@ def test_street
...
+
+  def test_complement
+    assert FFaker::AddressBR::COMPLEMENT.include?(FFaker::AddressBR.complement)
+  end
+
+  def test_neighborhood
+    neighborhood_prefix = FFaker::AddressBR.neighborhood.split(' ').first
+    assert FFaker::AddressBR::NEIGHBORHOOD_PREFIXES.include?(neighborhood_prefix)
+  end
+
```
Volosh1n, line 46: Could you also, please, provide test for full_address too (for both cases, with complement and without)? I know, there is no such test, but we can left this place better as it was before.

lib/ffaker/address_br.rb
```diff
@@ -37,9 +37,23 @@ def street
...
     end
+
+    def full_address(with_complement: false)
+      if with_complement
+        "#{street}, #{building_number}, #{complement}, #{neighborhood}, #{city}, #{state}, Brazil"
+      else
+      "#{street}, #{building_number}, #{neighborhood}, #{city}, #{state}, Brazil"
+      end
+    end
+
```
Volosh1n, line 57:
```suggestion
    def full_address(with_complement: false)
      base_address_order = [street, building_number, neighborhood, city, state, 'Brazil']
      return base_address_order.join(', ') unless with_complement
      
      base_address_order.insert(2, complement).join(', ')
    end
```
This is up to you

test/test_address_br.rb
az3vedo: Sure, I will have a look at it

lib/ffaker/address_br.rb
az3vedo: This is way better, thanks!

marocchino:
Thank you. The above review will be corrected in the main branch.

PR #514, opened 2021-12-03
Remove UniqueUtils from REFERENCE.md

Volosh1n:
`#is_a?` didn't filter out UniqueUtils, so we have `FFaker::UniqueUtils` in [REFERENCE.md](https://github.com/ffaker/ffaker/blob/main/REFERENCE.md#ffakeruniqueutils) 🤯 
Let's use `#instance_of?` instead.


scripts/reference.rb
```diff
@@ -11,24 +11,20 @@
...
-    next unless mod.is_a?(Module)
-    next if mod == FFaker::ArrayUtils
-    next if mod == FFaker::ModuleUtils
-    next if mod == FFaker::RandomUtils
-    next if mod == FFaker::Random
-
-    mod
-  end.compact
+  FFaker.constants
+    .select { |const| FFaker.const_get(const.to_s).instance_of?(Module) }
```
marocchino, line 20: @Volosh1n Why is `to_s` here?

scripts/reference.rb
```diff
@@ -11,24 +11,20 @@
...
+    .reject { |const| UTILS_MODULES.include?(const) }
+    .sort
```
marocchino, line 22: Unless you really want a symbol.
```suggestion
  FFaker.constants
    .map { |symbol| FFaker.const_get(const) }
    .select { |const| const.instance_of?(Module) }
    .reject { |const| UTILS_MODULES.include?(const) }
    .sort
```

marocchino (approved):

Issue #500, opened 2021-08-07
FFaker::HTMLIpsum.fancy_string can return a string with no HTML


xunker:
I have a case where, given the right random seed, `FFaker::HTMLIpsum.fancy_string` will return a string with no HTML.

Reproduction with v2.18.0 (ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin18]) (MacOS 10.15.7):
```ruby
> FFaker::Random.seed = 13 # Thirteen?!? Spooky unlucky coincidence!
> FFaker::Random.reset!
> str = FFaker::HTMLIpsum.fancy_string
# => "Placeat voluptatem alias iusto doloribus voluptatibus adipisci excepturi soluta. Nulla minima ex recusandae quisquam eius. Odit similique maxime sed quis vero aliquam quidem. Sunt ad ut officia nesciunt vitae earum sit deleniti. Esse tenetur perferendis fugit quasi. Reiciendis excepturi inventore ducimus harum numquam. Incidunt aperiam sint occaecati sunt aut. Neque accusantium deleniti provident voluptatibus doloribus. Eligendi doloribus pariatur itaque aspernatur similique. Aliquam enim cupiditate perspiciatis eius officia tempora cumque blanditiis. Error vel non harum sed." 
str.include?('<')
# => false 
```

All other methods won't have this issue regardless of the random seed used because they explicitly put in at least one HTML tag set via a `content_tag_for` call.

gabrielcosta42:
It's more common than it looks. The same output can be seen using a lot of other seeds (97, 176, 197).

gabrielcosta42:
Should be fixed after #516 PR.

PR #516, opened 2021-12-10
Ensure FFaker::HTMLIpsum.fancy_string return includes at least one tag

gabrielcosta42:
Issue #500 

- Non HTML paragraphs limited to `count - 1`, giving way to at least one tag.
- Added more other text formatting tags


marocchino:
Thanks!

PR #521, opened 2022-05-27
Fix crash in FFaker::Time#datetime when defining FFaker::Date module

brupla6126:
Defining the FFaker::Date module makes FFaker::Time#datetime crash. 

When migrating from Faker, we needed to add the module FFaker::Date to handle #backward and #forward differently than FFaker::Time.

FFaker::Time#datetime uses Date.new instead of ::Date.new so it crashes on trying to instantiate FFaker::Date which is a module.

### DIFF
--- a/test/test_time.rb
+++ b/test/test_time.rb
@@ -3,13 +3,6 @@
 require_relative 'helper'
 require 'date'
 
-module FFaker
-  module Date
-    extend FFaker::ModuleUtils
-    extend self
-  end
-end
-
 class TestFakerTime < Test::Unit::TestCase
   include DeterministicHelper
 
--- a/test/test_time.rb
+++ b/test/test_time.rb
@@ -3,6 +3,13 @@
 require_relative 'helper'
 require 'date'
 
+module FFaker
+  module Date
+    extend FFaker::ModuleUtils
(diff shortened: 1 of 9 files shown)


test/test_time.rb
```diff
@@ -3,6 +3,13 @@
 require_relative 'helper'
 require 'date'
 
+module FFaker
+  module Date
+    extend FFaker::ModuleUtils
+    extend self
```
Volosh1n, line 9: Shouldn't it live somewhere else? I mean, not in test folder

AlexWayfer: Maybe something like `::Date` is meaning here?

Volosh1n: I mean, I don't get the idea putting this module to test file. Overall I understand the issue

brupla6126: > Shouldn't it live somewhere else? I mean, not in test folder

I put the module there just to demonstrate the bug. I'll implement the FFaker::Date module with the methods we needed from our migration from Faker to FFaker.

Volosh1n (approved):
👍

marocchino (approved):
👍🏽

PR #522, opened 2022-05-27
Added FFaker::Internet.uuid method to make uuids

mnyon-grandkru:
Copied from [Faker](https://github.com/faker-ruby/faker/). Adds FFaker::Internet.uuid method. Includes a test

### DIFF
--- a/test/test_internet.rb
+++ b/test/test_internet.rb
@@ -129,4 +129,10 @@ class TestFakerInternet < Test::Unit::TestCase
     assert @tester.mac(nil).length == 12
     assert @tester.mac('').length == 12
   end
+
+  def test_uuid
+    uuid = @tester.uuid
+    assert_equal(36, uuid.size)
+    assert_match(/\A\h{8}-\h{4}-4\h{3}-\h{4}-\h{12}\z/, uuid)
+  end
 end
--- a/test/test_internet.rb
+++ b/test/test_internet.rb
@@ -9,7 +9,7 @@ class TestFakerInternet < Test::Unit::TestCase
     FFaker::Internet,
     :email, :free_email, :safe_email, :disposable_email,
     :user_name, :domain_name, :domain_word, :domain_suffix,
-    :http_url, :ip_v4_address, :password, :slug, :mac
+    :http_url, :ip_v4_address, :password, :slug, :mac, :uuid
   )
 
   def setup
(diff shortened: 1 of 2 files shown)


test/test_internet.rb
```diff
@@ -129,4 +129,10 @@ def test_mac_delimiter_argument
     assert @tester.mac(nil).length == 12
     assert @tester.mac('').length == 12
   end
+
+  def test_uuid
+    uuid = @tester.uuid
+    assert_equal(36, uuid.size)
+    assert_match(/\A\h{8}-\h{4}-4\h{3}-\h{4}-\h{12}\z/, uuid)
```
Volosh1n, line 136: Could you, please, also add #uuid method to `assert_methods_are_deterministic` in this file?

mnyon-grandkru: I will. I have some other work to do on this as well. Thanks for the feedback.

Volosh1n: Thank you too!

AlexWayfer:
`SecureRandom.uuid` seems shorten and simpler than `FFaker::Internet.uuid`. For what purposes is this PR?

Volosh1n:
`SecureRandom.uuid` is not deterministic, I think

AlexWayfer:
> `SecureRandom.uuid` is not deterministic, I think

You want a "secure" "random" "UUID"… the most one.

But OK, as you wish, I'm not against aliases.

Volosh1n:
@AlexWayfer Who said FFaker::Internet.uuid should be secure at all? Could you please help me find out where I said that? I replied to whole solution with SecureRandom, not to your comment at all.

mnyon-grandkru:
So the requirements for FFaker::Internet.uuid include generating a UUID in a non-deterministic manner. If it's not needed at all and SecureRandom.uuid will suffice for the purposes of creating a UUID then we don't need this PR at all and I can close this out.

Volosh1n:
> So the requirements for FFaker::Internet.uuid include generating a UUID in a non-deterministic manner. If it's not needed at all and SecureRandom.uuid will suffice for the purposes of creating a UUID then we don't need this PR at all and I can close this out.

Feel free to close, if there is no need in such method then. Thank you anyway

AlexWayfer:
> @AlexWayfer Who said FFaker::Internet.uuid should be secure at all? Could you please help me find out where I said that? I replied to whole solution with SecureRandom, not to your comment at all.

1. Nobody. But here is used a secured one solution.
2. You haven't said that. But here is still the same solution.
3. I understand, but:
    1. It's still shorten, what was my point.
    2. It's secure.
    3. It's still used by proposed code.

I hope you've got my point.

Volosh1n:
@AlexWayfer I think I got the idea. Thank you, no offense

PR #529, opened 2023-04-27
fix: Generate valid VINs

### LINKED ISSUE #310: FFaker::Vehicle.vin should generate valid vins
Currently the ones it generates usually don't pass the [check digit test](https://en.wikibooks.org/wiki/Vehicle_Identification_Numbers_(VIN_codes)/Check_digit).

philippevezina:
The `TRANSLITERATION_VALUES` were not mapped because the Hash's keys are strings but `.to_sym` was called for the key mappings. Letters "O" and "Q" were also in VINs at times because `VALID_WMI_REGIONS` could randomly sample these letters.

Fixes #310

### DIFF
--- a/lib/ffaker/vehicle.rb
+++ b/lib/ffaker/vehicle.rb
@@ -7,7 +7,7 @@ module FFaker
       extend self
 
       # https://en.wikibooks.org/wiki/Vehicle_Identification_Numbers_(VIN_codes)/World_Manufacturer_Identifier_(WMI)
-      VALID_WMI_REGIONS = [*'A'..'C', *'J'..'Z', *'1'..'9'].freeze
+      VALID_WMI_REGIONS = [*'A'..'C', *'J'..'N', 'P', *'R'..'Z', *'1'..'9'].freeze
 
       VALID_YEAR_CHARS = %w[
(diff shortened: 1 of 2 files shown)


test/test_vehicle.rb
```diff
@@ -5,6 +5,14 @@
 class TestVehicle < Test::Unit::TestCase
   include DeterministicHelper
 
+  # https://en.wikibooks.org/wiki/Vehicle_Identification_Numbers_(VIN_codes)/Check_digit
+  VIN_TRANSLITERATION_VALUES = {
```
masterT, line 9: Maybe support number as well, so there is a single mapping.

philippevezina: Yeah, I think we can keep this a simple bug fix for now :wink:

lib/ffaker/vehicle.rb
```diff
@@ -53,7 +53,7 @@ def vin
 
         # Calculate the Check Digit
         weighted_sum = generated_vin.chars.each_with_index.sum do |char, idx|
-          (TRANSLITERATION_VALUES[char.to_sym] || char).to_i * POSITION_WEIGHTS[idx]
```
marocchino, line 56: 🤔

PR #532, opened 2023-08-23
Added Ruby 3 fix for unique_utils

simonhildebrandt:
During our Ruby 3 upgrade I noticed a argument error while doing this:

```ruby
FFaker::Number.unique.number(digits: 1)
```

... turns out `FFaker::UniqueUtils#method_missing` doesn't handle keyword args the Ruby 3 way. This change (with test) fixes that.


AlexWayfer (approved):

AlexWayfer (changes requested):
I agree with proposed changes, but it breaks Ruby 2.5 and 2.6.

I suggest to drop their support: they're not supported officially even for security. 

The second way is to use `ruby2_keywords`: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#a-compatible-delegation
But I don't like that much.

marocchino:
I just drop under ruby 2.7.

AlexWayfer:
@simonhildebrandt rebase to the updated `main` please.

test/test_unique_numbers.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+class TestUniqueNumbers < Test::Unit::TestCase
+  def test_generates_unique_numbers
```
marocchino, line 6: We already have `test/test_unique_utils.rb`. Please remove this file and add to `test/test_unique_utils.rb`.

simonhildebrandt: I considered doing this, but since it's more about the *interaction* of these two behaviours it didn't seem like a good fit. Happy to move my test in there if you have a strong opinion about it though.

marocchino (approved):
LGTM, Thanks.

AlexWayfer (approved):

PR #533, opened 2023-08-23
Drop support under ruby 2.7

marocchino:


test/test_phone_number_sg.rb
```diff
@@ -46,7 +46,7 @@ def test_premium_service_number
 
   def test_phone_number
     10.times do
-      assert_match(/[6,8,9]\d{3}\s\d{4}/, @tester.phone_number)
+      assert_match(/[6,89]\d{3}\s\d{4}/, @tester.phone_number)
```
AlexWayfer, line 49: Is this intentional change?

AlexWayfer (commented):
I'm currently working on RuboCop offenses (complete) resolution, and maybe even `lint` in CI. So it'd help me a lot.

test/test_phone_number_sg.rb
AlexWayfer: Oh, it's chars in Regexp, haven't noticed. Probably then we should move `,` somewhere else, before or after digits.

Also it looks strange: phone number begins with `,`? Probably there should not be `,` at all.

marocchino: 2409bd026c894e777036415859c002e55ab10541

PR #535, opened 2023-08-29
Adds FFaker::Number.between

professor:

This adds a FFaker::Number.between(from: X, to: Y). These can be both integers or both floads.

### DIFF
--- a/lib/ffaker/number.rb
+++ b/lib/ffaker/number.rb
@@ -15,6 +15,10 @@ module FFaker
       FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
     end
 
+    def between(from: 1.00, to: 5000.00)
+      fetch_sample(from..to)
+    end
+
     private
 
     def generate_pattern(digits)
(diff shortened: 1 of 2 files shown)


lib/ffaker/number.rb
```diff
@@ -15,6 +15,10 @@ def decimal(whole_digits: 1, fractional_digits: 1)
       FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
     end
 
+    def between(from: 1.00, to: 5000.00)
+      rand(from..to)
```
Volosh1n, line 19: I believe it should be fetch_sample(from..to)’ since ‘rand’ does not use FFaker’s seed

test/test_number.rb
```diff
@@ -42,4 +42,16 @@ def test_decimal_when_invalid_argument
       @tester.decimal(fractional_digits: 0)
     end
   end
+
+  def test_between
```
Volosh1n, line 46: Please, add your method to assert_methods_are_deterministic at the top of this test file

lib/ffaker/number.rb
professor: Thank you for the tip. I'm making the fix.

test/test_number.rb
professor: Thank you for the reminder. Fix is coming.

Volosh1n (approved):

PR #537, opened 2023-08-29
Adds FFaker::Address.latitude and longitude

professor:

This adds a FFaker::Address.latitude and longitude.

### DIFF
--- a/lib/ffaker/address.rb
+++ b/lib/ffaker/address.rb
@@ -111,5 +111,13 @@ module FFaker
     def time_zone
       fetch_sample(TIME_ZONE)
     end
+
+    def latitude
+      fetch_sample(-90.0..90.0)
+    end
+
+    def longitude
+      fetch_sample(-180.0..180.0)
+    end
   end
 end
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
   - Add your change HERE
   - Update README [@professor]
+  - Added `FFaker::Address.latitude` and `FFaker::Address.longitude` [@professor]
   - Adds FFaker::Number.between [@professor]
   - Fixed `FFaker::Number.unique.number(digits: 1)` under Ruby 3 [@simonhildebrandt]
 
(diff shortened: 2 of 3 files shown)


lib/ffaker/address.rb
```diff
@@ -111,5 +111,14 @@ def country_code(given_country = nil)
     def time_zone
       fetch_sample(TIME_ZONE)
     end
+
+    def latitude
+      puts Module.nesting.inspect
```
Volosh1n, line 116: Is it needed or left after debug purpose?

lib/ffaker/address.rb
```diff
@@ -111,5 +111,14 @@ def country_code(given_country = nil)
...
+      Float((rand * 180) - 90)
```
Volosh1n, line 117: fetch_sample(-90.0..90.0)
Will it do the trick? Same for the second method

professor: I love it!

lib/ffaker/address.rb
professor: Thanks! Too many repos in front of me today.

lib/ffaker/address.rb
```diff
@@ -111,5 +111,13 @@ def country_code(given_country = nil)
     def time_zone
       fetch_sample(TIME_ZONE)
     end
+
+    def latitude
+      Float(fetch_sample(-90, 90))
```
marocchino, line 116:
```suggestion
      fetch_sample(-90.0..90.0)
```
will be fine.

marocchino:
It would be very helpful if you also update the `Changelog.md`.

lib/ffaker/address.rb
professor: Thank you for the suggestion. I'm making this change now. 

I would like latitude and longitude to be like `-130.11409542259364`

marocchino (approved):
LGTM

AlexWayfer:
Latitude and longitude are more something geographic, location, not address, which is named by aliases.

I've checked do we have such module — and yes, and there are `lat` and `lng` methods: https://github.com/ffaker/ffaker/blob/main/REFERENCE.md#ffakergeolocation

image

professor:
I'm closing this PR. I'd suggest to add an alias for `longitude` and `latitude` to make them easier to find.

PR #538, opened 2023-08-31
Adds FFaker::Number:leading_zero_number

professor:

This adds a FFaker::Number.leading_zero_number.

Note that this would return a string, not an integer. I'm not sure if that breaks any design constraints with FFaker (as compared to Faker). We only use it once, so I'm happy to inline the code in our app. I'm providing it here for anyone else transitioning from Faker to FFaker.

### DIFF
--- a/lib/ffaker/number.rb
+++ b/lib/ffaker/number.rb
@@ -15,6 +15,10 @@ module FFaker
       FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
     end
 
+    def leading_zero_number(digits: 10)
+      FFaker.numerify("0#{"#" * (digits - 1)}")
+    end
+
     def between(from: 1.00, to: 5000.00)
       fetch_sample(from..to)
     end
--- a/test/test_number.rb
+++ b/test/test_number.rb
@@ -5,7 +5,7 @@ require_relative 'helper'
 class TestNumber < Test::Unit::TestCase
   include DeterministicHelper
 
-  assert_methods_are_deterministic(FFaker::Number, :number, :decimal, :between)
+  assert_methods_are_deterministic(FFaker::Number, :number, :decimal, :between, :leading_zero_number)
 
   def setup
     @tester = FFaker::Number
@@ -41,6 +41,14 @@ class TestNumber < Test::Unit::TestCase
     assert_raise(ArgumentError.new('Digits cannot be less than 1')) do
       @tester.decimal(fractional_digits: 0)
     end
+  end
+
+  def test_leading_zero
+    assert @tester.leading_zero_number.is_a?(String)
+    assert @tester.leading_zero_number.length == 10
+    assert @tester.leading_zero_number(digits: 3).length == 3
+    assert_match(/\A0\d{9}\z/, @tester.leading_zero_number.to_s)
+    assert_match(/\A0\d{2}\z/, @tester.leading_zero_number(digits: 3).to_s)
   end
 
   def test_between
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 ## development
 
   - Add your change HERE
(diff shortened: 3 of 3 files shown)


lib/ffaker/number.rb
```diff
@@ -15,6 +15,10 @@ def decimal(whole_digits: 1, fractional_digits: 1)
       FFaker.numerify("#{whole_part_pattern}.#{fractional_part_pattern}").to_f
     end
 
+    def leading_zero_number(digits: 10)
+      "0#{FFaker::Number.number(digits: digits -1)}"
```
marocchino, line 19: how about? `"%0#{digits}d" % fetch_sample(0...10**(digits - 1))`

test/test_number.rb
```diff
@@ -43,6 +43,14 @@ def test_decimal_when_invalid_argument
     end
   end
 
+  def test_leading_zero
+    assert @tester.leading_zero_number.is_a?(String)
+    assert @tester.leading_zero_number.digits.count == 10
+    assert @tester.leading_zero_number(digits: 3).digits.count == 3
+    assert_match(/0\d/, @tester.leading_zero_number.to_s)
+    assert_match(/0\d{3}/, @tester.leading_zero_number(digits: 3).to_s)
```
marocchino, line 51:
```suggestion
    assert_match(/\A0\d{9}\z/, @tester.leading_zero_number.to_s)
    assert_match(/\A0\d{2}\z/, @tester.leading_zero_number(digits: 3).to_s)
```

lib/ffaker/number.rb
professor: Thanks!

marocchino (approved):

AlexWayfer (commented):
I don't understand the meaning of this method.

And I don't understand Faker motivation for:
* adding this;
* [deprecating this](https://github.com/faker-ruby/faker/pull/1516);
* [removing this](https://github.com/faker-ruby/faker/pull/510);
* [re-adding this](https://github.com/faker-ruby/faker/pull/1651).

Where are discussions, reasons?

I see this algorithm very strange and rare-case for me. What do you need it for? Why not a regular number? Why 10 digits is default? Why only one guaranteed zero?

marocchino (changes requested):
I realized you don't add :leading_zero_number to assert_methods_are_deterministic.
please add it?

professor:
> I realized you don't add :leading_zero_number to assert_methods_are_deterministic.
please add it?

Fixed

professor:
> I don't understand the meaning of this method.

In many systems, there are number (invoice numbers, account numbers) where there are leading 0's before the random numbers. 

My code base currently only uses this once and instead of typing
```
FFaker::Number.leading_zero_number(digits: 9)
```
I could replace it with
```
"0{FFaker::Number.number(digits: 8)}"
```

marocchino (approved):

AlexWayfer:
> In many systems, there are number (invoice numbers, account numbers) where there are leading 0's before the random numbers.

Yeah, I don't understand why it starts with 0. Why not with 1, 9, 111? It seems very specific for me, I've never seen or done such thing. I could only saw something like a company has ranges, and invoices are incrementing, and it looks like `10004567`, and can be `10876543` later.

So… if there any reason for single leading zero, standard — I'd like to know. Public reference. If no, if it's something specific and unexplained — I'm against a dedicate such method in library and for in-place simple implementations (just add 1 character).

For me it seems like "let's add a method, which returns Lorem Ipsum, but  in downcase, instead of external downcase method call".

I hope you understand.

AlexWayfer:
Oh… interesting. Yeah, I'm not familiar with this, being a younger man than these systems. If it's only the reason, I mean type of such data — then yes, I'm for not adding this as kind of outdated. It'd be strange to drop 4-years-old Ruby versions support, but support 30-years old data format, which can be implemented very simply by those who needed.

PR #534, opened 2023-08-23
Resolve RuboCop offenses

AlexWayfer:
And add RuboCop GitHub action.

Please note `.rubocop_todo.yml` config and changes to arguments.

PR #543, opened 2023-09-07
Adds FFaker::Bank.routing_number and accounting_number

professor:
This adds `FFaker::Bank.routing_number and accounting_number`

For our use case, we're looking for realistic looking US Bank account numbers and routing numbers. 
Accounting numbers can very between 9 and 17 digits [(Reference)](https://www.investopedia.com/articles/personal-finance/063015/routing-number-vs-account-number-how-they-differ.asp) Routing numbers have a checksum calculation on the last digit.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,8 @@
 ## development
 
   - Add your change HERE
+  - Add FFaker::BankUS.accounting_number [@professor]
+  - Add FFaker::BankUS.routing_number [@professor]
   - Resolve a lot of RuboCop offenses [@AlexWayfer]
   - Change `Image.url`, `Image.file`, `Book.cover` and `Avatar.image` arguments to keywords [@AlexWayfer]
   - Adds FFaker::Date.birthday [@professor]
--- a/lib/ffaker/bank_us.rb
+++ b/lib/ffaker/bank_us.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module FFaker
+  module BankUS
+    extend ModuleUtils
+    extend self
+
+    def account_number(min_digits: 9, max_digits: 17)
+      FFaker.numerify('#' * rand(min_digits..max_digits))
+    end
+
+    def routing_number
+      partial_routing_number = FFaker.numerify('########')
+      ninth_digit = generate_ninth_digit(partial_routing_number)
+
+      "#{partial_routing_number}#{ninth_digit}"
+    end
+
+    private
+
+    def generate_ninth_digit(num_string)
+      # This leverages the `Modules 10, Straight Summation` used for routing_numbers
+      # See http://www.sxlist.com/techref/ecommerce/bank/routingnumber/index.htm
+      # for more details
+      num_array = num_string.chars.map(&:to_i)
+      (
+        (7 * (num_array[0] + num_array[3] + num_array[6])) +
+          (3 * (num_array[1] + num_array[4] + num_array[7])) +
+          (9 * (num_array[2] + num_array[5]))
+      ) % 10
+    end
+  end
+end
--- a/test/test_bank_us.rb
+++ b/test/test_bank_us.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+class TestBankUS < Test::Unit::TestCase
+  include DeterministicHelper
+
+  assert_methods_are_deterministic(
+    FFaker::BankUS,
+    :account_number, :routing_number
+  )
+
+  def setup
+    @tester = FFaker::BankUS
+  end
+
+  def test_account_number
+    assert_instance_of String, @tester.account_number
+    assert_match(/\A\d{8}\z/, @tester.account_number(min_digits: 8, max_digits: 8))
+    assert_match(/\A\d{12}\z/, @tester.account_number(min_digits: 12, max_digits: 12))
+  end
+
+  def test_routing_number
+    routing_number = @tester.routing_number
+    assert_match(/\A\d{9}\z/, routing_number)
+
+    checksum = (
+      (7 * (routing_number[0].to_i + routing_number[3].to_i + routing_number[6].to_i)) +
+        (3 * (routing_number[1].to_i + routing_number[4].to_i + routing_number[7].to_i)) +
+        (9 * (routing_number[2].to_i + routing_number[5].to_i + routing_number[8].to_i))
+    )
+
+    assert_equal(0, checksum % 10, 'The routing number\'s checksum is not a multiple of ten')
+  end
+end


lib/ffaker/bank.rb
```diff
@@ -114,5 +125,14 @@ def check_country_existence(country_code)
...
     end
+
+    def checksum(num_string)
+      num_array = num_string.split('').map(&:to_i)
+      (
+        7 * (num_array[0] + num_array[3] + num_array[6]) +
+          3 * (num_array[1] + num_array[4] + num_array[7]) +
+          9 * (num_array[2] + num_array[5])
+      ) % 10
+    end
```
AlexWayfer, line 136: Is it a common checksum method, or specific one? I mean… there are MD5, HMAC, etc., as I know. Probably we should be more specific with the method name. Up to `routing_number_checksum`.

lib/ffaker/bank.rb
```diff
@@ -103,6 +103,17 @@ def card_type
...
 
+    def account_number(min_digits: 9 , max_digits: 17)
+      FFaker.numerify('#' * rand(min_digits..max_digits))
+    end
+
+    def routing_number
+      partial_routing_number = FFaker.numerify('########')
+      checksum = checksum(partial_routing_number)
+
+      "#{partial_routing_number}#{checksum}"
```
AlexWayfer, line 114: Honestly, I'd prefer interpolation of method call here. We don't need for a variable — it's not re-using. But it can be a part of taste.

lib/ffaker/bank.rb
```diff
@@ -114,5 +125,14 @@ def check_country_existence(country_code)
 
       raise ArgumentError, "Unexpected country code: '#{country_code}'"
     end
+
+    def checksum(num_string)
+      num_array = num_string.split('').map(&:to_i)
```
AlexWayfer, line 130: We don't have integration with RuboCop yet (#534), but I believe you should use [`String#chars`](https://docs.rubocop.org/rubocop/cops_style.html#stylestringchars) here.

test/test_bank.rb
```diff
@@ -50,4 +50,22 @@ def test_card_expiry_date
...
+  def test_routing_number
+    routing_number = @tester.routing_number
+    checksum = (
+      7 * (routing_number[0].to_i + routing_number[3].to_i + routing_number[6].to_i) +
+        3 * (routing_number[1].to_i + routing_number[4].to_i + routing_number[7].to_i) +
+        9 * (routing_number[2].to_i + routing_number[5].to_i + routing_number[8].to_i)
+    ) % 10
+
+    assert_match(/\d{9}/, routing_number)
+    assert_equal(0, checksum)
```
AlexWayfer, line 69: Is it kind of… reversible checksum? I mean, you can take a result, do the same math operation and get zero? Sorry if the question is stupid.

AlexWayfer (commented):
I'm not against such methods.

But, we probably should think about `US` suffix somewhere (in the module name?), if it's specific to a country.

test/test_bank.rb
```diff
@@ -50,4 +50,22 @@ def test_card_expiry_date
   def test_card_type
     assert_include @tester::CARD_TYPES, @tester.card_type
   end
+
+  def test_account_number
+    assert_instance_of String, @tester.account_number
+    assert_match(/\d{8}/, @tester.account_number(min_digits: 8, max_digits: 8))
+    assert_match(/\d{12}/, @tester.account_number(min_digits: 12, max_digits: 12))
```
marocchino, line 57: I think we always have to test whole string.
```suggestion
    assert_match(/\A\d{8}\z/, @tester.account_number(min_digits: 8, max_digits: 8))
    assert_match(/\A\d{12}\z/, @tester.account_number(min_digits: 12, max_digits: 12))
```

test/test_bank.rb
marocchino: maybe)
how about..
```rb
assert(valid_routing_number?(routing_number))

# somewhere below..
def valid_routing_number?(routing_number)
  ((
    7 * (routing_number[0].to_i + routing_number[3].to_i + routing_number[6].to_i) +
    3 * (routing_number[1].to_i + routing_number[4].to_i + routing_number[7].to_i) +
    9 * (routing_number[2].to_i + routing_number[5].to_i + routing_number[8].to_i)
  ) % 10) == 0
end
```

lib/ffaker/bank.rb
professor: In general, I find code that separates the presentation logic from the calculation logic easier to read. I agree that the local variables are barely used. 

So yes, this feels like a style difference.

lib/ffaker/bank.rb
professor: Nice!

test/test_bank.rb
professor: The two checksum algorithms (the one in the implementation and the one in the test) are slightly different. The implementation uses generates the last digit of the routing number. The test verifies that the routing number is valid. See part 2 of https://www.wikihow.com/Calculate-the-Check-Digit-of-a-Routing-Number-from-an-Illegible-Check for an example.

In the test case, we want to verify that the verification number is a multiple of 10.

test/test_bank.rb
professor: Thank you. I've incorporated this change.

lib/ffaker/bank.rb
AlexWayfer: OK, you have the point, I have no problems with a simple local variable. :upside_down_face:

lib/ffaker/bank.rb
professor: Looks like it is called the `Modules 10, Straight Summation` method. The algorithm was created in 1911 -- I would expect MD5, etc to have different design considerations and thus not be a good fit. See more here: http://www.sxlist.com/techref/ecommerce/bank/routingnumber/index.htm

professor:
> But, we probably should think about US suffix somewhere (in the module name?), if it's specific to a country.

Done.

test/test_bank.rb
AlexWayfer: > In the test case, we want to verify that the verification number is a multiple of 10.

Why?

It seems for me that you should test the last digit of a generated `routing_number`, not properties of internal checksum.

lib/ffaker/bank.rb
AlexWayfer: Okay…

It'd be nice to leave the link as a comment to the code, do you agree?

professor: I agree. I'm adding it as a comment.

test/test_bank.rb
professor: I'm not understanding your original concern. Perhaps we can schedule a zoom call. I'm in California and I'm at UTC - 7 hours.

> It seems for me that you should test the last digit of a generated routing_number, not properties of internal checksum.

I'm taking a different approach than either listed. I would like the ruby test to verify that the generated routing number is valid. To do that, we follow the algorithm described [here]( http://www.sxlist.com/techref/ecommerce/bank/routingnumber/index.htm#Check) and other places. The algorithm multiples each digit by 1 or 7 and then adds the final digit. The sum of the products should be a multiple of 10.  

This means that this test is not testing the last digit of the generated routing number (which would be bad since the test is simply duplicating the code that its testing and nothing is then tested), but rather verifying that the routing number is valid.

Part of the problem is that the word "checksum" is overloaded here. In this example from the website, the routing number `067803457` would generate a sum of products of 130. I would call 130 the checksum, but some might call 7 the checksum, since that's the digit that verifies the rest are digits are correct. I'm not sure who is really correct and I'm not sure that it matters.

Since you last look at the code, I have modified it and the error message slightly to make this clearer. In the test, I'm calling the sum of products the `checksum` and in the implementation, I'm calling the code that generates the last digit `generate_ninth_digit`

Changelog.md
```diff
@@ -1,6 +1,8 @@
 ## development
 
   - Add your change HERE
+  - Add FFaker::Bank.accounting_number [@professor]
+  - Add FFaker::Bank.routing_number [@professor]
```
marocchino, line 5:
```suggestion
  - Add FFaker::BankUS.accounting_number [@professor]
  - Add FFaker::BankUS.routing_number [@professor]
```

professor: Thank you

marocchino (approved):
👍🏽

professor:
@marocchino thank you for the approval. This PR has been in a green state for a little over 7 days with no new comments. I would love to see it merged this week.

marocchino:
You've waited too long. I think this will be okay.

PR #546, opened 2023-10-29
Add UUID module

stilist:
I recently used `FFaker::Guid.guid` and realized it can produce invalid UUIDs: UUIDs can’t be arbitrary data; bits 48–51 and 64–65 must have predefined values. Rather than change `Guid.guid` to produce a valid UUID, I’ve deprecated it and replaced it with more-specific `UUID.uuidv4`, `.uuidv6`, `.uuidv7`, and `.uuidv8` methods. These follow [the specification](https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/) (which is still being worked on, but close to approval — [current draft](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-13.txt)).

* I chose not to implement UUID v1, v2, v3, or v5 because I don’t remember seeing them used in the past 15 years. They can be added to the module if needed, though.
* I debated whether to have the v6 and v7 methods use real timestamps by accepting `from`, `to` arguments like `Date.between`. I decided it wasn’t necessary, though it might be useful for testing code that relies on those versions’ behavior of putting the slowest-changing part of the timestamp at the start of the UUID. (For example, a trie.)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,8 @@
 ## development
 
   - Add your change HERE
+  - Add `FFaker::UUID` `.uuidv4`, `.uuidv6`, `.uuidv7`, and `.uuidv8` [@stilist]
+  - Deprecate `FFaker::GUID` in favor of `FFaker::UUID` [@stilist]
   - Limit FFaker::BankUS.routing_number first two digits [@professor]
 
 # 2.23.0
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
   - Add your change HERE
   - Add `FFaker::UUID` `.uuidv4`, `.uuidv6`, `.uuidv7`, and `.uuidv8` [@stilist]
-  - Deprecate `FFaker::GUID` in favor of `FFaker::UUID` [@stilist]
+  - Deprecate `FFaker::Guid.guid` in favor of `FFaker::UUID` methods [@stilist]
   - Limit FFaker::BankUS.routing_number first two digits [@professor]
 
 # 2.23.0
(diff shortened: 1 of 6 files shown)


lib/ffaker/guid.rb
```diff
@@ -5,7 +5,13 @@ module Guid
...
     extend self
 
+    # Because this method uses arbitrary hexadecimal characters it is likely to
+    # generate invalid UUIDs--UUIDs must have a version (1-8) at bits 48-51,
+    # and bits 64-65 must be 0b10.
+    #
+    # @deprecated Often generates invalid UUIDs. Use {UUID} instead.
     def guid
+      warn '[guid] is deprecated. Use the UUID module instead.'
       FFaker.hexify('########-####-####-####-############')
```
marocchino, line 15: Anyone who wanted to create an incorrect uuid will be in trouble... But, I think it will be okay.
```suggestion
      warn '[guid] is deprecated. Use the UUID.uuidv4 method instead.'
      FFaker::UUID.uuidv4
```

Changelog.md
```diff
@@ -1,6 +1,8 @@
 ## development
 
   - Add your change HERE
+  - Add `FFaker::UUID` `.uuidv4`, `.uuidv6`, `.uuidv7`, and `.uuidv8` [@stilist]
+  - Deprecate `FFaker::Guid.guid` in favor of `FFaker::UUID` methods [@stilist]
```
marocchino, line 5: It would be better if there was a notice about lowercase strings.

stilist: Decided to call `.upcase` on `Guid.guid` instead to avoid the breaking change. 👍

PR #548, opened 2024-03-28
Fix Luhn algorithm used in french companie's SIRET and sweden SSN

Krap:
Fixed implementation for the Luhn checksum algorithm (implementation was wrong, and examples in reference where all invalid).


marocchino (approved):
LGTM

Volosh1n (approved):

marocchino:
Thanks!

PR #551, opened 2025-01-14
Generator for Finnish identity numbers

zHarrowed:
Small pr adding support for generating Finnish identity numbers
https://en.wikipedia.org/wiki/National_identification_number#Finland

### DIFF
--- a/lib/ffaker/identification_fi.rb
+++ b/lib/ffaker/identification_fi.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module FFaker
+  module IdentificationFI
+    extend ModuleUtils
+    extend self
+
+    CHECK_DIGITS = "0123456789ABCDEFHJKLMNPRSTUVWXY".freeze
+    # Number ranges in real use
+    POSSIBLY_REAL_FEMALE_INDIVIDUAL_NUMBERS = (2..898).step(2).to_a.freeze
+    POSSIBLY_REAL_MALE_INDIVIDUAL_NUMBERS = (3..899).step(2).to_a.freeze
+    # Number ranges not in real use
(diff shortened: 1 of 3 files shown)


lib/ffaker/identification_fi.rb
```diff
@@ -0,0 +1,71 @@
...
+      month = fetch_formatted_month(birthday)
+      year = fetch_formatted_year(birthday)
+      separator = fetch_separator(birthday)
+      individual_number = fetch_individual_number(gender, fake)
+      check_digit = calculate_check_digit(birthday, individual_number)
+      "#{day}#{month}#{year}#{separator}#{individual_number}#{check_digit}"
+    end
+
+    private
+      def fetch_formatted_day(birthday)
```
marocchino, line 27: We don't indent one more deps for private methods. could you please lint it?

lib/ffaker/identification_fi.rb
```diff
@@ -0,0 +1,71 @@
...
+        sprintf('%.3d', fetch_sample(numbers_range))
+      end
+
+      def calculate_check_digit(birthday, individual_number)
+        digit = "#{birthday.day}#{fetch_formatted_month(birthday)}#{fetch_formatted_year(birthday)}#{individual_number}".to_i % 31
+        CHECK_DIGITS[digit]
+      end
+
+      def check_birth_year(birth_year)
+        return if birth_year.between?(1799, 2100)
```
marocchino, line 66: If you have a reason for setting the upper limit to the year 2100, I would appreciate it if you could share it.

lib/ffaker/identification_fi.rb
zHarrowed: Sure, ran Rubocop and those fixes now included :+1:

marocchino:
I'm not in a position to work on it right now, but I'll merge it tomorrow. Thank you for your patience.

PR #553, opened 2025-03-28
feat: add support for `FFaker::Boolean.boolean(true_ratio:)`

nedzib:
This adds the FFaker::Boolean.boolean(true_ratio:) method.

In our use case, we need to generate seed data automatically. For some fields, it's important to bias the result toward true to better reflect real usage scenarios.

Example usage:
FFaker::Boolean.boolean(true_ratio: 0.9) → returns true ~90% of the time.

Default ratio is 0.5.

Tests included in test/test_boolean.rb.
image

### DIFF
--- a/test/test_boolean.rb
+++ b/test/test_boolean.rb
@@ -5,10 +5,41 @@ require_relative 'helper'
 class TestBoolean < Test::Unit::TestCase
   include DeterministicHelper
 
-  assert_methods_are_deterministic(FFaker::Boolean, :maybe)
+  assert_methods_are_deterministic(FFaker::Boolean, :maybe, :boolean)
 
   def test_maybe
     maybe = FFaker::Boolean.maybe
     assert [true, false].include?(maybe)
+  end
+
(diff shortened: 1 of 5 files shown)


marocchino (approved):
LGTM

test/test_boolean.rb
```diff
@@ -5,10 +5,41 @@
...
+  def test_boolean_with_true_ratio
+    true_ratio = 0.8
+    true_count = 0
+    1000.times do
+      true_count += 1 if FFaker::Boolean.boolean(true_ratio: true_ratio)
+    end
+    assert_in_delta true_ratio, true_count / 1000.0, 0.1
+  end
+
+  def test_boolean_with_true_ratio_0
```
marocchino, line 32:
```suggestion
  def test_boolean_with_true_ratio_zero
```
or
```suggestion
  def test_boolean_with_true_ratio0
```

nedzib: done

